PR #5, opened 2026-06-03
Restore server player animators

Garward:
## Summary

Restores vanilla server side entity shape loading and reconnects player animators on join when they were initialized before loaded server shapes were available.

This fixes server side animation frame callback NRE spam during held item actions while preserving animation timing for mods that rely on authoritative animation callbacks.

It also fixes a decompiler/regeneration issue where the animation/tag network packet classes lost vanilla protobuf implicit field contracts. Without that, Stratum serialized `BulkAnimationPacket` payloads as `bytes=0`, so stock clients received empty entity animation packets and mobs slid around without animations.

## Type

- [x] Bug fix
- [x] Performance
- [ ] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

- [x] `.\scripts\extract-patches.sh` ran clean.
- [x] Confirmed after restoring `ProtoContract(ImplicitFields = ImplicitFields.AllFields)`, mobs animate again on an unmodded client.

 ## Notes

  This avoids the broad `LoadEntityShapes(server.EntityTypes, server.api)` approach. Player shapes are loaded up front, while non player entity shapes/animators are loaded on
  demand when server side animation behavior requires them.

### DIFF
--- a/patches/VintagestoryApi/Common/Model/Animation/AnimationManager.cs.patch
+++ b/patches/VintagestoryApi/Common/Model/Animation/AnimationManager.cs.patch
@@ -1,0 +1,358 @@
+diff --git a/VintagestoryApi/Common/Model/Animation/AnimationManager.cs b/VintagestoryApi/Common/Model/Animation/AnimationManager.cs
+index 573ec12..4d57a87 100644
+--- a/VintagestoryApi/Common/Model/Animation/AnimationManager.cs
++++ b/VintagestoryApi/Common/Model/Animation/AnimationManager.cs
+@@ -27,11 +27,23 @@ namespace Vintagestory.API.Common
+         public bool AnimationsDirty { get; set; }
+ 
+         /// <summary>
+         /// The animator for the animation manager.
+         /// </summary>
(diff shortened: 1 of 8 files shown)


Garward:
ah mobs aren't animating anymore gotta test this further nvm

trevorftp:
No worries, that confirms this needs more testing. The player animator fix is probably valid, but we should try to separate two cases:

1. Player server animators: needed for held item/server callback behavior.
2. Mob/global server animators: potentially needed for some mods, but expensive if restored for everything.

I’d avoid merging the broad `LoadEntityShapes(server.EntityTypes, server.api)` version until we know whether mobs actually need server animators for gameplay/mod callbacks, and what the tick cost is under load

trevorftp (approved):
looks good to me.

commit is scoped to one logical fix and matches the title: restore server player animators after vanilla server shape loading comes back. the added join hook + helper guard rails are clean, and it follows stratum style (`Stratum` prefix + clear marker comment).

no blockers from me, approved.

PR #6, opened 2026-06-04
Restore decompiled serialization metadata

Garward:
## Summary

Restores vanilla Newtonsoft serialization metadata that the decompiler failed to decode.

Several attributes were emitted as `/*Could not decode attribute arguments.*/`, which changed rebuilt runtime metadata from vanilla. This restores `JsonObject(MemberSerialization.OptIn)` and the missing `JsonProperty` name/null handling arguments so Stratum preserves vanilla JSON serialization behavior for settings, server config/player data, macros, and glTF metadata.

## Type

- [x] Bug fix
- [ ] Performance
- [ ] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

- [x] `.\scripts\extract-patches.sh` ran clean.

  ## Testing

- Compared rebuilt serialization metadata against vanilla assemblies locally; `VintagestoryLib` had `0` remaining vanilla serialization metadata drift after the fix. (at least from what I found)
- Rebuilt the Linux single file server binary and windows binary.
- Ran the rebuilt Linux binary with `--stratum-refresh`; it applied the overlay and reached `Dedicated Server now running`.

### DIFF
--- a/patches/VintagestoryLib/Vintagestory.Client.NoObf/ClientSettings.cs.patch
+++ b/patches/VintagestoryLib/Vintagestory.Client.NoObf/ClientSettings.cs.patch
@@ -1,0 +1,17 @@
+diff --git a/VintagestoryLib/Vintagestory.Client.NoObf/ClientSettings.cs b/VintagestoryLib/Vintagestory.Client.NoObf/ClientSettings.cs
+index 9786197..c1e2c75 100644
+--- a/VintagestoryLib/Vintagestory.Client.NoObf/ClientSettings.cs
++++ b/VintagestoryLib/Vintagestory.Client.NoObf/ClientSettings.cs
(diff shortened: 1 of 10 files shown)


trevorftp (approved):
yeah this looks good to me!! commit does what it says on the tin: restores the busted decompiler json metadata instead of changing runtime behavior.

big win is it removes the Could not decode attribute arguments noise and puts explicit serialization intent back (MemberSerialization.OptIn, proper json property args), so this is safer and easier to reason about going forward.

**non-blocking future idea:** _we should probably add_ a tiny decompile post-process step (or metadata rehydrate pass) so these known attribute-arg decode failures get fixed automatically in bootstrap/extract, instead of relying on manual cleanup commits. not needed for this PR though.

no blockers from me, ship it.

PR #24, opened 2026-06-15
Optimize RegisterBlock method

tehtelev:
## Summary

We're reducing the number of pre-allocation calls for the registered block array by a factor of 100. Block is a heavy structure, so raising this value too high doesn't make sense.

## Type

- [ ] Bug fix
- [x] Performance
- [ ] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

## Performance numbers
- Reduction in function time: from 423 ms to 231 ms;
- Reduction in GC mem allocations for Block[]: from 1.74 GB to 0.197 GB.

**Before**
registerblock до trace
registerblock до mem 1
registerblock до mem 2

**After**
registerblock после trace
registerblock после mem
registerblock после mem 2


trevorftp (approved):
Solid change the benchmarks look good. LGTM 👍

PR #25, opened 2026-06-16
Optimize ModRegistryObjectTypeLoader as much as possible

tehtelev:
## Summary

The object type loading system was refactored, resulting in the following key improvements:

1. Efficient parallel execution: Instead of launching multiple threads with manual task distribution, **Parallel.ForEach** with **Partitioner** is now used, ensuring automatic load balancing and optimal utilization of CPU cores.
2. Lazy JSON resolution: The need for expensive **DeepClone()** for each object variant has been eliminated. Instead, a static Resolve method has been introduced that processes only those parts of the JSON that require modification, significantly reducing memory consumption and speeding up variant processing.
3. Improved code structure: The logic is divided into clear steps-loading base types, processing them in parallel, and final registration. This improves readability and makes the code easier to maintain.
4. Reliable thread management: The **_gatheringCompleted** flag has been introduced, eliminating fragile wait loops via **Thread.Sleep**, making execution more predictable and efficient.
5. Enhanced error handling: errors in processing certain types no longer interrupt the entire process, but are isolated and logged, ensuring that other objects continue loading.

[This is the implementation of my PR for the Vintage story repository](https://github.com/anegostudios/vsessentialsmod/pull/42)

## Type

- [ ] Bug fix
- [x] Performance
- [ ] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

## Performance numbers
- Reduction in function AssetLoaded time: from 14.1 s to 13.1 s;
- Reduction in function GatherVariantsAndPopulate summary time: from 45.0 s to 21.9 s;
- Reduction in GC mem allocations summary: from 23.28 GB to 16.51 GB.

**Before**
modregistry до trace
modregistry до mem
modregistry до mem 2

**After**
modregistry после trace
modregistry после mem
modregistry после mem 2

### DIFF
--- a/patches/VSEssentials/Loading/RegistryObjectType.cs.patch
+++ b/patches/VSEssentials/Loading/RegistryObjectType.cs.patch
@@ -1,0 +1,247 @@
+diff --git a/VSEssentials/Loading/RegistryObjectType.cs b/VSEssentials/Loading/RegistryObjectType.cs
+index 7c78fc7..fa5a8a5 100644
+--- a/VSEssentials/Loading/RegistryObjectType.cs
++++ b/VSEssentials/Loading/RegistryObjectType.cs
+@@ -276,19 +276,11 @@ namespace Vintagestory.ServerMods.NoObf
+                 Enabled = Enabled,
+                 jsonObject = jobject,
+                 Variant = variant
+             };
+ 
+-            try
+-            {
+-                solveByType(jobject, fullcode.Path, variant);
+-            }
+-            catch (Exception e)
+-            {
+-                api.Server.Logger.Error("Exception thrown while trying to resolve *byType properties of type {0}, variant {1}. Will ignore most of the attributes. Exception thrown:", this.Code, fullcode);
+-                api.Server.Logger.Error(e);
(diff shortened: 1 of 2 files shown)


trevorftp (approved):
Code looks good moving from in place mutation to a copy-on-write approach was the right fix,,,

One thing to potentially keep in mind down the line? `NeedsResolve` and `Resolve` traverse the same nodes twice on anything that actually needs processing, so there’s still room to combine them into a single pass if startup time becomes a target again. Not a blocker at all given the numbers. LGTM 👍

tehtelev:
> Code looks good moving from in place mutation to a copy-on-write approach was the right fix,,,
> 
> One thing to potentially keep in mind down the line? `NeedsResolve` and `Resolve` traverse the same nodes twice on anything that actually needs processing, so there’s still room to combine them into a single pass if startup time becomes a target again. Not a blocker at all given the numbers. LGTM 👍

Since this entire structure works recursively (there's no other way), the only way to speed it up is to rewrite the entire game from Newtonsoft Json to System.Text.Json

PR #27, opened 2026-06-18
Optimize WildcardUtil.Match

tehtelev:
## Summary

Key changes:
- Removed dependency on Regex for standard substitutions;
- Faster path for exact matches.
- Consolidated processing via the fastMatch method.
- Variants safety check for the presence of the * character before calling the variant extraction logic.

## Type

- [ ] Bug fix
- [x] Performance
- [ ] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

## Performance numbers

- Reduction in function time: from 1491 ms to 166 ms;
- Reduction in GC mem allocations: from 564 MB to 14 MB.

**Before**
wildcard до trace
wildcard до mem

**After**
wildcard после trace
wildcard после mem

### DIFF
--- a/patches/VintagestoryApi/Util/WildcardUtil.cs.patch
+++ b/patches/VintagestoryApi/Util/WildcardUtil.cs.patch
@@ -1,0 +1,100 @@
+diff --git a/VintagestoryApi/Util/WildcardUtil.cs b/VintagestoryApi/Util/WildcardUtil.cs
+index 64fb462..cafac88 100644
+--- a/VintagestoryApi/Util/WildcardUtil.cs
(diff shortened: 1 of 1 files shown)


trevorftp (approved):
The regex removal is the big one here!! calling `Regex.IsMatch` with a freshly constructed pattern string on every variant match was always going to be rough at scale, and the numbers show it!! Holy. The null guard at the top is also a real correctness fix, the old code would’ve thrown if either argument came in null since it hit `.Domain` before any null check. Variant guard skipping `MatchesVariants` when there’s no `*` in the path is a nice touch too. Logic ordering is solid, cheapest checks first. LGTM 👍

PR #33, opened 2026-06-21
Optimize DiscDepositGenerator

tehtelev:
## Summary

As noted in the source code, using Perlin noise to impart non-uniformity to ore deposits is quite excessive (and slow). I propose replacing it with a faster hash function that also produces smooth noise values ​​from 0 to 1, but is calculated instantly. Visually, the ore deposits are barely distinguishable from the original ones. You can also mix in a world seed for greater uniqueness, if needed.
During testing, I regenerated chunks around me with the command **/wgen regen 10**. 

A comparison of several points produced by noise generators. The blue dots indicate the curve from the source code. The orange color is the curve of the function I proposed.
image

Generation comparison. [Block Overlay](https://mods.vintagestory.at/xray) mod was used to view ores.

A refined implementation of this [PR](https://github.com/anegostudios/vssurvivalmod/pull/170)

## Type

- [ ] Bug fix
- [x] Performance
- [ ] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

## Performance numbers

Testing on 10,200 generation chunks:
- Total method execution time: decreased from 32.6 s to 9.1 s;
 - GC load: decreased from 256 MB to 198 MB.
 
Total chunk generation time decreased from 2m 20s to 2m 12s (-5.7%).

**Before**
disc before trace
disc before mem
1  before
2 before
3  before

**After**
disc after trace
disc after nen
1 after
2 after
3 after

### DIFF
--- a/patches/VSEssentials/Systems/WorldGen/Standard/ChunkGen/5.GenDeposits/Generators/DiscGenerator.cs.patch
+++ b/patches/VSEssentials/Systems/WorldGen/Standard/ChunkGen/5.GenDeposits/Generators/DiscGenerator.cs.patch
@@ -1,0 +1,198 @@
+diff --git a/VSEssentials/Systems/WorldGen/Standard/ChunkGen/5.GenDeposits/Generators/DiscGenerator.cs b/VSEssentials/Systems/WorldGen/Standard/ChunkGen/5.GenDeposits/Generators/DiscGenerator.cs
+index 2b6633c..90ad42e 100644
+--- a/VSEssentials/Systems/WorldGen/Standard/ChunkGen/5.GenDeposits/Generators/DiscGenerator.cs
++++ b/VSEssentials/Systems/WorldGen/Standard/ChunkGen/5.GenDeposits/Generators/DiscGenerator.cs
+@@ -206,21 +206,56 @@ namespace Vintagestory.ServerMods
+             var rnd = new LCGRandom(Api.World.Seed);
+             absAvgQuantity = GetAbsAvgQuantity(rnd);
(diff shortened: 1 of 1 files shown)


trevorftp (changes requested):
the numbers are solid and most of this looks good. `FastNoise` is a clean Wang hash variant, the `numx`/`numz` extraction in `loadYPosAndThickness` is correct, no issues there.

---

**Bug: `posyChunksizeMod` / `posyChunksizeDiv` are stale after the first y-iteration**

These are cached before the y-loop:

```cs
int posyChunksizeMod = posy % chunksize;
int posyChunksizeDiv = posy / chunksize;

for (int y = 0; y < hereThickness; y++)
{
    ...
    int index3d = ((posyChunksizeMod) * chunksize + lz) * chunksize + lx;
    int blockId = chunks[posyChunksizeDiv].Data.GetBlockIdUnsafe(index3d);
    ...
    posy--;  // cached values are now wrong for the next iteration
}
```

`posy--` runs every iteration, so y=0 is correct and everything after reads/writes the wrong chunk layer and block index. Any deposit with `hereThickness > 1` (most of them) would have all but the first layer misplaced. Fix is just moving the two declarations inside the loop right above `index3d`,, you still eliminate the duplicate division calls the original had per iteration.

tehtelev:
Yeah. That's where the problem was. I'll fix it tomorrow then.

tehtelev:
Fixed. Also updated screenshots.
I can also make a statistical comparison of the amount of generated ores before and after, if needed.

PR #37, opened 2026-06-22
Add config toggles for connection chat messages

### LINKED ISSUE #17: Add config toggles for join, leave, and crash chat messages
### Problem

Some servers want quieter public chat. Join, leave, and crash messages can be useful, but they can also spam chat during restarts, events, or crash loops.

### Proposed change

Add Stratum config for player connection messages.

Add separate toggles for:

- player join messages
- player leave messages
- player crash/disconnect error messages

Keep defaults close to vanilla unless there is a good reason not to.

Also research whether these can be moved to another built-in chat tab, like an info/system tab, instead of only showing in general chat. Do not depend on a custom client for this.

### Alternatives considered

_No response_

### Scope

Zaldaryon:
Three new booleans in the Chat config section let server admins suppress join, leave, and disconnect/kick messages from public chat. ShowJoinMessages, ShowLeaveMessages, and ShowDisconnectMessages all default to true so nothing changes on upgrade.

The server log (Logger.Event) still records every connection event. Suppressing only affects what players see in chat. Useful for servers that restart often, run events with high player churn, or want quieter general chat during peak hours.

The issue also asked about routing messages to a separate chat tab. The built-in ServerInfoChatGroup (-1) only renders for players with controlserver privilege, so routing there would hide messages from normal players without feedback. Leaving that for a follow-up if there is a workable approach.

Closes #17

### DIFF
--- a/patches/VintagestoryLib/Vintagestory.Server/ServerMain.cs.patch
+++ b/patches/VintagestoryLib/Vintagestory.Server/ServerMain.cs.patch
@@ -79,7 +79,10 @@ index 1017be9..bb2d35a 100644
  		client.State = EnumClientState.Connected;
  		client.MillisecsAtConnect = totalUnpausedTime.ElapsedMilliseconds;
 -		joinLeaveDeathMessage(Lang.Get("player-joined", player.PlayerName), EnumChatType.JoinLeave, player);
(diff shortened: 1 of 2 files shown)


trevorftp (approved):
Simple, effective, send it!

PR #38, opened 2026-06-22
Add /stratum pathfinding report command

### LINKED ISSUE #13: Add pathfinding queue metrics and report output
### Problem

Pathfinding can be a real source of pressure, but we do not have a good report for it. If mobs stop moving or the server starts lagging, it is hard to tell if the queue is full, workers are busy, or stale tasks are getting dropped.

### Proposed change

Add pathfinding metrics and a /stratum pathfinding report.

Report:

- queued tasks
- active worker tasks
- completed tasks
- dropped stale tasks
- dropped dead-owner tasks
- dropped oldest tasks
- average queue wait
- average solve time
- worker count
- queue high-water mark

Also show pathfinding pressure in /stratum doctor or /stratum performance.

### Alternatives considered

Use timings only. Shows cost, not queue health.

### Scope

Zaldaryon:
Exposes the queue metrics that PathfindingAsync already tracks but does not surface anywhere. The new /stratum pathfinding command shows worker count, current queue depth against the configured cap, priority mode state, and per-category drop counters (overflow, stale, far-entity).

The report method (StratumBuildReport) lives in the VSEssentials patch. The command handler in CmdStratum calls it via reflection since VintagestoryLib cannot reference VSEssentials at compile time. Same cross-assembly pattern that StartStratumExtraWorkers uses to read pathfinding config. Reflection cost is one invocation per admin command, zero impact on the pathfinding hot path.

Not included in this PR: solve-time tracking (needs a Stopwatch around AStar.FindPath in ProcessQueue) and queue high-water mark. Both need new instrumentation and can come as a follow-up once the baseline report is in place.

Closes #13

### DIFF
--- a/patches/VSEssentials/Entity/Pathfinding/PathfindingAsync.cs.patch
+++ b/patches/VSEssentials/Entity/Pathfinding/PathfindingAsync.cs.patch
@@ -370,6 +370,19 @@ index 48a886b..7b443ba 100644
 +            task.Finished = true;
 +        }
 +
++        // Stratum: expose queue metrics for /stratum pathfinding.
++        public string StratumBuildReport()
++        {
++            return string.Join("\n",
++                "Workers=" + (1 + stratumExtraWorkers.Count),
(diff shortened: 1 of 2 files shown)


trevorftp (approved):
clean, send it!

PR #39, opened 2026-06-22
Add /stratum doctor pressure summary

### LINKED ISSUE #15: Add /stratum doctor pressure summary
### Problem

There are a bunch of Stratum reports now. During lag, you still need one command that says what looks bad right now and where to look next.

### Proposed change

Add `/stratum doctor`.

Show a compact pressure summary:

- recent tick time/TPS
- chunk send pressure
- chunk generation pressure
- packet queue/back-pressure state
- pathfinding queue pressure
- autosave delay/flush state
- event listener pressure
- entity tick pressure
- join queue/join stage pressure
- pregen pressure if active

Example output lines:

"Chunk generation queue is high; inspect /stratum chunks."
"Pathfinding queue is near limit; inspect /stratum pathfinding."
"Autosave has delayed for 80s due to chunk pressure."

### Alternatives considered

Keep separate commands only. Slower during an incident.

### Scope

Zaldaryon:
Single command that shows what looks bad right now during lag. Runs through each subsystem and only prints a line when that subsystem reports pressure. Each line includes the relevant /stratum sub-command to inspect next.

Checks: tick time vs configured budget, packet back-pressure (deferred count and queue depth this tick), chunk send pressure (SkippedOutboundPressure from the send path), chunk generation pressure (clients with generation deferred), entity throttle state (PreviousTickOverloaded and throttle count), block listener skips, pathfinding queue depth against its configured cap (via reflection on PathfindingAsync.StratumBuildReport, graceful no-op when unavailable or queue is uncapped), autosave delays, pregen pause state, and connection queue depth.

All thresholds are data-driven. The doctor fires on the same indicators that each subsystem uses internally to decide to throttle, skip, or defer. No hardcoded magic numbers. When no subsystem reports pressure, the output is a single "no pressure detected" line.

Adds a DoctorSnapshot() method to StratumPerformanceStats that returns the private last-tick fields the doctor needs (autosave delayed, entities throttled, skipped outbound pressure, block listeners skipped, generation deferred clients) under the existing lock.

Closes #15

### DIFF
--- a/sources/VintagestoryLib/Vintagestory.Server/CmdStratum.cs
+++ b/sources/VintagestoryLib/Vintagestory.Server/CmdStratum.cs
@@ -22,7 +22,7 @@ internal class CmdStratum
 		server.api.commandapi.Create(StratumInfo.Id)
 			.WithAlias("serverinfo")
 			.WithDesc("Show Stratum server information")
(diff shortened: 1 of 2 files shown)


trevorftp:
Clean, conflicts fixed. Send it

PR #40, opened 2026-06-22
Add /stratum regions diagnostics command

### LINKED ISSUE #9: Add region ticking diagnostics
### Problem

**`Performance.RegionTicking`** is experimental. If someone turns it on, they need to see what it is actually doing. Right now it is hard to tell if it helped, did nothing, or is throwing problems in the background.

### Proposed change

Add `/stratum regions`.

Report:

- whether region ticking is enabled
- region size and worker count
- total entity regions this tick/window
- parallel regions vs serial tail regions
- entities per region
- slowest regions by tick time
- exception count and recent sample
- fallback status if automatic fallback is added later

Also print a startup/preflight warning when this is enabled

### Alternatives considered

_No response_

### Scope

Zaldaryon:
Fixes #9

Adds `/stratum regions` to report the state of the experimental region ticking system. Operators who enable `Performance.RegionTicking` can see whether the parallelism produces the expected region split, how entities distribute across regions, and whether exceptions accumulate.

Reports config (enabled, region size, workers, threshold), per-tick region/entity counts (parallel vs serial), the largest region by entity count, cumulative exception count with the last sample, and a placeholder for future automatic fallback status. Prints a startup warning when region ticking is enabled so operators notice the experimental flag in server logs.

Defers per-region tick timing to a follow-up. Entity count proxies for slowest region until Stopwatch overhead in the parallel path gets profiled.

### DIFF
--- a/patches/VintagestoryLib/Vintagestory.Server/ServerSystemEntitySimulation.cs.patch
+++ b/patches/VintagestoryLib/Vintagestory.Server/ServerSystemEntitySimulation.cs.patch
@@ -1,13 +1,14 @@
 diff --git a/VintagestoryLib/Vintagestory.Server/ServerSystemEntitySimulation.cs b/VintagestoryLib/Vintagestory.Server/ServerSystemEntitySimulation.cs
-index 750557a..337640d 100644
+index 750557a..5265f5f 100644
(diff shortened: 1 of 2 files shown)


trevorftp:
Conflicts fixed, looks good. Send it

PR #41, opened 2026-06-22
Add /stratum timings listeners with per-listener attribution

### LINKED ISSUE #14: Add event listener attribution for slow callbacks
### Problem

We can see event tick cost, but not always who caused it. If a listener is slow, server owners need the mod/system name or at least a useful callback name.

### Proposed change

Track slow block/entity listeners with better names.

Report:

- listener/profiler name
- owning mod or system if detectable
- listener interval
- average time over rolling window
- worst sample over rolling window
- call count
- whether adaptive throttling affected it

Expose through /stratum timings listeners and include top offenders in /stratum doctor.

### Alternatives considered

Aggregate timing only. Does not tell you what to fix.

### Scope

Zaldaryon:
Fixes #14

Adds `/stratum timings listeners` to list all registered entity and block tick listeners grouped by owning type. For each type the command reports registration count, configured tick intervals, and (when timings are running) the average duration, peak duration, total call count, and slow call count from the StratumTimings slow-listener buckets.

Also reports adaptive throttle configuration and current overload state so operators can see which listeners are being stretched and by how much.

Includes a `BuildSlowListenersDoctorLine` helper that produces a one-line worst-offender summary. Once #39 merges, wiring this into `HandleDoctor` is a single call.

The timing window covers the period since the last `/stratum timings start` or `/stratum timings reset`. Operators control the window size on demand. A fixed circular buffer per listener would add per-tick bookkeeping across thousands of registered callbacks for marginal benefit.

No patches to vanilla files. Both changed files live in `sources/`.

### DIFF
--- a/sources/VintagestoryLib/Vintagestory.Server/CmdStratum.cs
+++ b/sources/VintagestoryLib/Vintagestory.Server/CmdStratum.cs
@@ -9,6 +9,7 @@ using Vintagestory.API.Common.Entities;
 using Vintagestory.API.Config;
 using Vintagestory.API.MathTools;
 using Vintagestory.API.Server;
+using Vintagestory.Common;
 
 namespace Vintagestory.Server;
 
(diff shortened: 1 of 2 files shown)


sources/VintagestoryLib/Vintagestory.Server/CmdStratum.cs
```diff
@@ -445,14 +446,137 @@ private TextCommandResult HandleTimings(string action)
...
+
+		return TextCommandResult.Success(output.ToString());
+	}
+
+	// Stratum start: slow listeners doctor summary (#14)
+	internal static string BuildSlowListenersDoctorLine(
+		List<(string Name, long Calls, double TotalMs, double MaxMs, long SlowCalls)> slowEntity,
+		List<(string Name, long Calls, double TotalMs, double MaxMs, long SlowCalls)> slowBlock)
+	{
+		int totalSlow = 0;
```
trevorftp:
```suggestion
		long totalSlow = 0;
```

sources/VintagestoryLib/Vintagestory.Server/CmdStratum.cs
```diff
@@ -445,14 +446,137 @@ private TextCommandResult HandleTimings(string action)
...
+		string worstName = null;
+		double worstMax = 0;
+		foreach (var entry in slowEntity)
+		{
+			totalSlow += (int)entry.SlowCalls;
```
trevorftp: `SlowCalls` is a long but gets cast into an int.
```suggestion
			totalSlow += entry.SlowCalls;
```

trevorftp (changes requested):
Looks good overall couple small changes id recommend, others posted on their lines but, the listener totals currently use `entityListeners.Count` and `blockListeners.Count`, but those lists can contain null spots until pruning runs. Since the grouping loops skip null listeners, id assume the totals can show more listeners than are actually active.

potential fix:

```csharp
int activeEntityListeners = entityListeners.Count(listener => listener != null);
int activeBlockListeners = blockListeners.Count(listener => listener != null);
```

Zaldaryon:
@trevorftp done

trevorftp (approved):
Looks good!

PR #43, opened 2026-06-22
Pin ilspycmd version via tool manifest

### LINKED ISSUE #42: Pin ilspycmd version to prevent patch drift
Bootstrap installs ilspycmd via `dotnet tool install -g ilspycmd` without a version constraint. Different contributors (and CI) get different versions over time, which produces subtly different decompiler output for the same DLL.

The practical result: patches generated against one ilspycmd version fail to apply on a machine running a different version. The mismatch shows up as single-character differences in ILSpy comments (e.g. `constrained. prefix` vs `.constrained prefix`) that shift context lines and break hunk matching.

Current failure: `ClientPlatformWindows.cs.patch` fails on every bootstrap when ilspycmd 10.0.0.8330 decompiles the same DLL that patches were generated against a prior version.

Proposed fix: add a `.config/dotnet-tools.json` manifest that pins the exact ilspycmd version, and update both bootstrap scripts to restore from the manifest instead of installing globally. This guarantees every contributor and CI produces identical decompiler output.

Zaldaryon:
Fixes #42

Add `.config/dotnet-tools.json` pinning ilspycmd to 10.1.0.8386. Both bootstrap scripts read the pinned version from the manifest and install that exact version globally. Falls back to latest if the manifest is missing.

Without a pinned version, different ilspycmd releases produce subtly different decompiler output for the same DLL. Patches generated against one version fail to apply on machines running another. The `ClientPlatformWindows.cs` patch broke on every bootstrap using 10.0.x because of a single-character difference in an ILSpy comment (`constrained. prefix` vs `.constrained prefix`).

Tested on Linux (bash path). The PowerShell path uses the same logic via `ConvertFrom-Json` which ships with PowerShell 5.1+ on all Windows 10+ machines. Both scripts still fall back to latest when the manifest file is absent so forks that strip it keep working.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -1,0 +1,12 @@
+{
+  "version": 1,
+  "isRoot": true,
+  "tools": {
+    "ilspycmd": {
+      "version": "10.1.0.8386",
+      "commands": [
+        "ilspycmd"
+      ]
+    }
+  }
+}
--- a/scripts/bootstrap.ps1
+++ b/scripts/bootstrap.ps1
@@ -112,8 +112,16 @@ try {
     }
 
     if (-not (Get-Command ilspycmd -ErrorAction SilentlyContinue)) {
-        Write-Host "Installing ilspycmd"
-        dotnet tool install -g ilspycmd | Out-Null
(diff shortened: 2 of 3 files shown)


trevorftp:
Absolutely great catch, send it!

PR #56, opened 2026-06-23
Skip per-tick JSON deserialization in jail enforcement

Zaldaryon:
EnforceJailedPlayers called TryGetActiveJail for every online client every 250ms. TryGetActiveJail deserializes the full StratumJailStatus JSON (3.7 KB allocation) regardless of Active state. The jail key persists after release, so every ever-jailed player paid this cost forever: 25 us and 3,768 bytes per player per tick, all discarded.

At 50 ever-jailed players online: 1.25 ms/tick (2.5% of 50ms budget), 44 MB/min allocation rate, 5 Gen0 GCs/min from this path alone. At 200: 5 ms/tick (10% budget).

Fix: maintain a HashSet<string> of actively-jailed UIDs in memory. Populated on join (ApplyJailOnJoin already loads the record once), added on /jail, removed on /unjail, cleared on disconnect. EnforceJailedPlayers checks the set (O(1), zero alloc) instead of deserializing per client.

Tested: server boots, jail/unjail works, enforcement teleports back, clean shutdown. Measured with a standalone benchmark that calls DeserializeObject<StratumJailStatus> in a loop and reports per-call allocation and timing.


trevorftp:
Resolved, sent! Good stuff

Issue #10, opened 2026-06-05
Add region ticking allowlist, denylist, and automatic fallback


trevorftp:
### Problem

Some entities are not safe to tick on worker threads. Mods can touch shared world state, block accessors, spawn/despawn, or random shared static state. One bad entity type can make the whole region ticking mode risky

### Proposed change

Add allow/block controls and a fallback.

Suggested config:

```
Performance.RegionTicking.AllowedEntityCodePrefixes
Performance.RegionTicking.BlockedEntityCodePrefixes
Performance.RegionTicking.BlockedEntityClasses
Performance.RegionTicking.FallbackAfterExceptions
Performance.RegionTicking.FallbackWindowSeconds
```

Behavior:

- players stay on the main thread
- blocked entities tick on the main thread
- allowed entities can tick in region workers
- repeated exceptions turn region ticking off until reload or fallback to main-thread ticking
- `/stratum regions` reports fallback reason

### Scope

- [x] This is a server-side change. It does not require a custom client.

Zaldaryon:
I can take this. Here is the design I have in mind before writing code.

The bucketing loop in TickEntitiesRegionParallel currently splits players vs non-players. The filter would add a third destination: a serialBlockedList that ticks on main thread after the parallel pass, same as the serial tail for small regions. Checking entity code against the blocked prefixes uses the same StartsWith + array scan pattern that AmbientFaunaCodePrefixes already uses in ShouldThrottleStratumEntityTick.

Config shape:

```
RegionTicking.EntityFilterMode: "all" | "denylist" | "allowlist"
RegionTicking.BlockedEntityCodePrefixes: ["trader-", "humanoid-"]
RegionTicking.BlockedEntityClasses: ["EntityTrader"]
RegionTicking.AllowedEntityCodePrefixes: []
RegionTicking.FallbackAfterExceptions: 5
RegionTicking.FallbackWindowSeconds: 60
```

In denylist mode, any entity whose Code.Path starts with a blocked prefix or whose Properties.Class matches a blocked class goes to the serial list. In allowlist mode, only entities matching the allowed prefixes go parallel. Default mode "all" keeps current behavior.

Automatic fallback: a per-entity-type exception counter (ConcurrentDictionary<string, int>) increments on catch in TickOne. When a type exceeds FallbackAfterExceptions within FallbackWindowSeconds, the runtime adds that type to an in-memory blocked set for the session. /stratum regions reports the runtime-blocked types and the triggering exception count.

[...]

PR #59, opened 2026-06-24
Add region ticking entity filter and automatic fallback

Zaldaryon:
Implements #10.

Three filter modes for `Performance.RegionTicking`:

- `EntityFilterMode = "all"` (default): current behavior, all non-players go parallel.
- `EntityFilterMode = "denylist"`: entities matching `BlockedEntityCodePrefixes` or `BlockedEntityClasses` tick on the main thread. Everything else goes parallel.
- `EntityFilterMode = "allowlist"`: only entities matching `AllowedEntityCodePrefixes` go parallel. Everything else stays serial.

`BlockedEntityClasses` matches against `Properties.Class` (the string from the entity JSON, per your comment on #10).

Automatic fallback: a per-class exception counter (`ConcurrentDictionary<string, (count, windowStart)>`) increments on each parallel tick exception. When a type hits `FallbackAfterExceptions` (default 5) within `FallbackWindowSeconds` (default 60), the runtime adds it to a blocked set for the session. `/stratum reload` clears the set by default. `PersistFallbackUntilRestart = true` keeps it until restart.

`/stratum regions` reports the filter mode, how many entities the filter forced serial, and which classes got runtime-blocked.

Config shape:

```
[Performance.RegionTicking]
EntityFilterMode = "denylist"
BlockedEntityCodePrefixes = ["trader-", "humanoid-"]
BlockedEntityClasses = ["EntityTrader"]
AllowedEntityCodePrefixes = []
FallbackAfterExceptions = 5
FallbackWindowSeconds = 60
PersistFallbackUntilRestart = false
```

The filter runs once per entity at bucketing time (before the parallel pass). Cost: one prefix scan per non-player entity per tick, same as the existing ambient fauna check.

### DIFF
--- a/patches/VintagestoryLib/Vintagestory.Server/ServerSystemEntitySimulation.cs.patch
+++ b/patches/VintagestoryLib/Vintagestory.Server/ServerSystemEntitySimulation.cs.patch
@@ -1,5 +1,5 @@
 diff --git a/VintagestoryLib/Vintagestory.Server/ServerSystemEntitySimulation.cs b/VintagestoryLib/Vintagestory.Server/ServerSystemEntitySimulation.cs
-index 750557a..5265f5f 100644
+index 750557a..3d902a0 100644
 --- a/VintagestoryLib/Vintagestory.Server/ServerSystemEntitySimulation.cs
 +++ b/VintagestoryLib/Vintagestory.Server/ServerSystemEntitySimulation.cs
 @@ -1,8 +1,12 @@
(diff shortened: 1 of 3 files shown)


Zaldaryon:
This branch predates the tick-state-reconcile (#58) and expand-timings-profiler (#44) merges. Both landed on ServerSystemEntitySimulation.cs.patch after this branch forked from main. A rebase will conflict in that file. PR #60 fixes corrupt hunk headers in the current main version of the patch, so rebase after that one lands.

PR #61, opened 2026-06-25
Skip registering disabled Stratum commands

Zaldaryon:
## Summary

Stratum registered its player and staff commands unconditionally, so running it next to a mod that owns the same names crashed that mod at startup. Th3Essentials is the common case. This gates each command on the config entry that already governs it, so an operator can disable a Stratum command and let the other mod take the name.

## Problem

Command registration happens in OnBeginConfiguration and never checks whether the operator turned a command off. When Th3Essentials calls Create("msg") and Stratum already registered /msg, the API throws "Command with such name already exists". That exception propagates out of Th3Essentials.StartServerSide and the whole mod fails to load, so /spawn, /home, /back and the rest silently stop working. There was no way to resolve the conflict from config.

## Change

Add StratumCommandRegistration.ShouldRegister and wrap every player and staff command registration in it. ShouldRegister reads the Commands.* entry that already exists for that command, returns false when the group is disabled, and logs the skip with the config path so the reason is visible in server-main.log. Disabling Commands.Spawn, Commands.Back, Commands.Message or Commands.Homes.* in stratum-commands.json leaves those names free for the other mod. The serverinfo alias is also removed from /stratum so Th3Essentials can own /serverinfo.

## Testing

Built Release and booted a throwaway world with Th3Essentials 2.17.1, joined, and shut down cleanly.
- Default config: /stratum answers from Stratum and /serverinfo from Th3Essentials, with no duplicate-command error at boot.
- Conflict config: disabled Commands.Spawn/Back/Message/Homes.* in stratum-commands.json and enabled the Th3Essentials equivalents. /spawn, /back, /msg, /r, /home, /home set and /home del answer from Th3Essentials, while unrelated Stratum commands /near and /seen keep working. The startup crash is gone (127 systems start, 0 errors), and each gated command logs its skip line.

## Impact

Defaults do not change. Every group ships enabled, so a server with no conflicting mod registers the same commands as before. The only visible difference without a config edit is the removed /serverinfo alias. Disabled commands are fully unregistered rather than hidden, so a name like /sethome disappears entirely when its group is off, which is the intended result when handing it to another mod.

## Decision points

1. Bootstrap dependency. This branch needs #60 to bootstrap, since main still carries the corrupt hunk headers in the entity-simulation patch. No overlap in files, but #60 should land first.
2. Mod loader resolver. While testing I found that ModAssemblyLoader.LoadAssemblyDefinition lost its Cecil assembly resolver in decompilation, which makes the built lib fail to resolve VintagestoryAPI when scanning a mod in Mods for ModSystems. It is unrelated to this change and affects any mod, so I left it out. I can open it as its own PR with the restored ReaderParameters.
3. serverinfo alias. I dropped it unconditionally rather than only when another mod claims the name. Unconditional is predictable and matches the other gates, but I can make it conditional if you prefer.

## Docs

The command config in stratum-commands.json is not documented in the README. If useful, I can add a short section on disabling a Stratum command to defer the name to another mod. Left out here to keep the PR focused.

### DIFF
--- a/sources/VintagestoryLib/Vintagestory.Server/CmdStratum.cs
+++ b/sources/VintagestoryLib/Vintagestory.Server/CmdStratum.cs
@@ -21,7 +21,6 @@ internal class CmdStratum
 	{
 		this.server = server;
 		server.api.commandapi.Create(StratumInfo.Id)
-			.WithAlias("serverinfo")
 			.WithDesc("Show Stratum server information")
 			.WithArgs(server.api.commandapi.Parsers.OptionalWord("status|version|update|health|reload|preflight|packets|performance|perf|timings|players|player|chunks|entities|queues|pathfinding|doctor|regions|violations|access|chat|pregen|get|set|save"), server.api.commandapi.Parsers.OptionalWord("argument"), server.api.commandapi.Parsers.OptionalWord("detail"), server.api.commandapi.Parsers.OptionalWord("value1"), server.api.commandapi.Parsers.OptionalWord("value2"), server.api.commandapi.Parsers.OptionalWord("value3"), server.api.commandapi.Parsers.OptionalWord("value4"))
 			.RequiresPrivilege(Privilege.controlserver)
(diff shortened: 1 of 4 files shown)


trevorftp (approved):
Looks good, I like the solution.

PR #77, opened 2026-07-01
Refine NewNormalizedSimplexFractalNoise

tehtelev:
## Summary

Here are the main code changes that have virtually eliminated GC garbage generation:

- **Replacing local allocations with a thread-static cache.** Instead of creating new arrays (OctaveEntry[], PastEvaluation[], double[] maxValues, int[] order) in the constructor of each ColumnNoise, static fields with the [ThreadStatic] attribute are now used. Arrays are allocated once per thread and reused for all subsequent calls.

- **Lazy initialization and cache expansion.** When first accessed from a thread, arrays are created with a size of Max(48, number of octaves). If more octaves are needed in the future, the arrays are automatically expanded. This covers all real-world configurations without unnecessary allocations.

- **Fixing the actual number of used octaves (nUsedOctaves).** The nUsedOctaves field has been added, storing the exact number of active octaves. All loops (NoiseSign, Noise) now iterate up to this limit, rather than up to the array length. This allows for safe use of larger reusable arrays.

- **Resetting state via overwriting.** Data from previous evaluations (pastEvaluations) is overwritten in place. Unused elements are simply not read due to the loop limit of nUsedOctaves.

[This PR has also been sent to the game developers.](https://github.com/anegostudios/vsapi/pull/84)

## Type

- [ ] Bug fix
- [x] Performance
- [ ] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

## Performance numbers

Testing on 10,200 generation column chunks:
- GC load: decreased from 10.22 Gb to 250 MB

Comparisons of execution times are impossible because these modifications are performed in parallel threads, which the OS may manage differently each time. The only noticeable difference is the GC runtime.

**Before**
newnormalize before mem
newnormalize before trace

**After**
newnormalize after mem
newnormalize after trace


patches/VintagestoryApi/Math/Noise/NewNormalizedSimplexFractalNoise.cs.patch
```diff
@@ -0,0 +1,165 @@
...
+         public struct ColumnNoise
+         {
+-            OctaveEntry[] orderedOctaveEntries;
+-            PastEvaluation[] pastEvaluations;
++            // Stratum start:
++            // Testing on 10,200 generation column chunks:
++            // GC load: decreased from 10.22 Gb to 250 MB
++
++            // Thread-isolated static arrays (TLS).
++            // Size 48 covers any possible world height.
```
trevorftp, line 21: Is this meant to be octave count instead of world height?

PR #104, opened 2026-07-04
Implemented combat cheat detection and improved reach checks

trevorftp:
## Summary

Added ability to detect and prevent combat related cheats, as well as improving detection for outside of block placement and breaking. Such as interacting with entities, doors, tools, etc.

## Type

- [ ] Bug fix
- [ ] Performance
- [x] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

### DIFF
--- a/patches/VintagestoryLib/Vintagestory.Server/ServerSystemEntitySimulation.cs.patch
+++ b/patches/VintagestoryLib/Vintagestory.Server/ServerSystemEntitySimulation.cs.patch
@@ -1,5 +1,5 @@
 diff --git a/VintagestoryLib/Vintagestory.Server/ServerSystemEntitySimulation.cs b/VintagestoryLib/Vintagestory.Server/ServerSystemEntitySimulation.cs
-index 750557a..d7950fd 100644
+index 750557a..5cb19d7 100644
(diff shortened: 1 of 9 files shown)


trevorftp:
Merging into dev, review still appreciated but will test for some time.

Zaldaryon:
> Merging into dev, review still appreciated but will test for some time.

I found two issues worth fixing before this sits in indev for long:

1. `StratumCombatGuard.TryAcceptAttack` has no call site. The PR adds the combat guard and reporter path, but no patch wires it into `HandleEntityInteraction` before the attack reaches `TriggerPlayerInteractEntity` or `entity.OnInteract`. As a result, the aura and aim-cone checks never record violations or cancel hits.
2. `NoFall` and `MovementAuthority` sit behind the old movement gate in `StratumMovementGuard.TryAccept`. If `Anticheat.Movement.Enabled` is false, or both `DetectFlight` and `DetectNoclip` are false, the method returns before the new checks run. That makes the new config sections impossible to use on their own.

Small cleanup: `git diff --check` reports trailing whitespace in: 
- `StratumAnticheatConfig.cs` and
- `StratumBlockBreakGuard.cs`.

PR #142, opened 2026-07-12
Add recurring dropped item cleanup task

### LINKED ISSUE #102: USER SUGGESTION: Automatic ground item cleanup
### Problem

Automatic ground clearing of items and related entities, with ability to adjust time, if it announces to chat and also the exact chat message. 

### Proposed change

n/a

### Alternatives considered

_No response_

### Scope

xeljel:
## Summary

Add a scheduled item cleanup for loose items on the ground (or in the water).

Features:
- Configurable cleanup interval
- Configurable minimum item age
- Configurable warning, start, and finish messages
- Configurable warning times

## Type

- [ ] Bug fix
- [ ] Performance
- [x] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

## Related issues

Closes #102

### DIFF
--- a/StratumServer/stratum.default.json
+++ b/StratumServer/stratum.default.json
@@ -238,6 +238,15 @@
       "OverloadDropThresholdMs": 500.0,
       "FloorChunks": 4,
       "AdjustmentIntervalSeconds": 2.0
+    },
+    "ItemCleanup": {
+      "Enabled": false,
+      "IntervalSeconds": 60,
+      "MinimumEntityAgeSeconds": 10,
+      "CleanupWarningMessage": "Cleaning up dropped items in {0} seconds.",
+      "CleanupWarningTimeOffsets": [30, 10],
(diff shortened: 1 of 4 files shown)


trevorftp:
Woohoo congrats on your first commit! Glad to get some help :)

two things I would like to see fixed before a merge though, 
`MinimumEntityAgeSeconds` is being compared to `server.ElapsedMilliseconds`. so that means the default 10 is 10ms instead of 10s, that needs multiplied x1000.

also the cleanup tasks calls, `Entity.Die` from the thread pool, I would keep the actual despawn on the main server thread, by queueing the entity ids back to the main thread or by doing the cleanup from the scheduled callback directly. Entity state/chunk/network despawn flow is not something we should try to mutate from a worker thread

and a minor thing is doing a git diff shows some trailing whitespaces in `StratumItemCleanup.cs`

PR #155, opened 2026-07-14
Fixed issue with windows timer overshooting with `Sleep()`

trevorftp:
## Summary

This fix combines the 1ms Windows timer request with a deadline-based tick wait, so the server sleeps once for most of the remaining budget and only yields at the very edge instead of relying on one overshooting `Sleep()`

## Type

- [ ] Bug fix
- [ ] Performance
- [ ] New feature
- [x] Refactor or cleanup
- [ ] Docs or build

## Checklist

### DIFF
--- a/patches/VintagestoryLib/Vintagestory.Server/ServerMain.cs.patch
+++ b/patches/VintagestoryLib/Vintagestory.Server/ServerMain.cs.patch
@@ -501,7 +501,7 @@ index 1017be9..0ab2741 100644
  			{
  				processConnectionQueue();
  				statsupdate = DateTime.UtcNow;
-@@ -1533,18 +1726,25 @@ public sealed class ServerMain : GameMain, IServerWorldAccessor, IWorldAccessor,
+@@ -1533,22 +1726,29 @@ public sealed class ServerMain : GameMain, IServerWorldAccessor, IWorldAccessor,
  				{
  					StatsCollector[StatsCollectorIndex].tickTimes[k] = 0L;
  				}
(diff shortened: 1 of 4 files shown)


Zaldaryon:
Tested this locally: patches apply clean, build is 0 errors and 0 warnings on the touched files, and a full boot through the real self-extracting launcher (`-p:EmbedPatchedDlls=true`, so the patched `VintagestoryLib.dll` is actually loaded, not vanilla) reaches RunGame and ticks without errors.

Found one issue in `WaitForNextTick`. When `FineSleepGranularity` is false, Windows timer resolution disabled in config or `timeBeginPeriod` failed, the loop spins the last 17ms of every tick through `Thread.Yield()`/`SpinWait()` instead of sleeping. I reproduced the wait loop standalone and measured CPU time consumed per tick: the old single `Thread.Sleep()` cost 0.4% of a core, this PR's fine-granularity path costs 3.8%, and the no-fine-granularity fallback costs 50.5%. That pegs half a core every tick for anyone who turns off the timer resolution feature on purpose, which is the whole reason that config option exists, or anyone whose `timeBeginPeriod` call fails.

Pushed a fix (8b6424c): the fallback now does one coarse `Thread.Sleep()` for the whole remaining budget instead of spin-waiting, back to the old behavior for that case. Re-measured after the fix, the fallback path drops to 0.2% CPU and the fine-granularity path is unaffected (3.5%, same order as before). Verified again with a real server boot after the fix, reaches RunGame and shuts down clean.

PR #172, opened 2026-07-19
Build the server assets packet synchronously at boot

### LINKED ISSUE #151: Process-killing race: 'Collection was modified' in ItemTypeNet.GetItemTypePacket during background BuildServerAssetsPacket
**Versions:** Stratum v1.22.3-stratum.15 (linux-x64 release zip), base game 1.22.3. Observed twice in CI (GitHub Actions ubuntu-latest, 4 cores), not reproducible on a 24-core desktop.

**What happened.** The background server assets packet build crashed the whole process with an unhandled exception on a thread-pool worker:

```
[Fatal] Failed network encoding block game:spearhead-chert:
[Error] Exception: Collection was modified; enumeration operation may not execute.
   at Vintagestory.Common.ItemTypeNet.GetItemTypePacket(Item item, IClassRegistryAPI registry, FastMemoryStream ms) in ItemTypeNet.cs:line 230
   at Vintagestory.Server.ServerMain.BuildServerAssetsPacket() in ServerMain.cs:line 5101
   at Vintagestory.API.Common.TyronThreadPool.<>c__DisplayClass11_0.<QueueTask>b__0(Object _)
```

Being unhandled on a pool thread, it kills the server process outright.

**Context and window.** This came out of the StratumParity differential suite (#147): servers are booted headless in-process, a mod is staged (its compilation lengthens the boot), and the scenario joins a player and performs bulk world edits right after launch. On a slow 4-core runner that activity overlaps the off-thread `BuildServerAssetsPacket` enumeration of every block and item, and the enumeration hits a collection being mutated concurrently. Two consecutive CI runs crashed at the same place; the vanilla leg of the same matrix passed both times, which may be timing luck (the involved frames are vanilla-derived code, this area carries no Stratum patch) or may point at a Stratum-side lazy cache mutating item state at first use. I could not tell from the logs which collection was being mutated.

**Why it matters beyond tests.** The same window exists on production servers with heavy modpacks: the assets packet build runs while the server is already live, and early player activity (or mod init work) that mutates item/block collections concurrently can kill the whole server with no recovery path, since the exception escapes on a pool thread.

**Workaround used in the suite:** delay bulk world mutation until the assets packet reports built (bounded grace period otherwise).

**Repro material.** Full server logs of the crashing boots are attached as CI artifacts on the failed runs of https://github.com/Pixnop/StratumParity/pull/8 (workflow run 29289846080, artifact server-logs-stratum). Happy to help narrow it down; if you can confirm it on stock 1.22.3 binaries, it probably belongs upstream at Anego's tracker.

Pixnop:
## Summary

Fixes #151: the process-killing race in the background server-assets-packet build.

**Mechanism, proven at line level** (decompiled with the repo's pinned ilspycmd, so lines match the shipped PDBs): `StartBuildServerAssetsPacket()` queues `BuildServerAssetsPacket` on TyronThreadPool between WorldReady and worldgen startup. The encoders it calls enumerate shared `ClassRegistry` dictionaries through LINQ (`ItemTypeNet.GetItemTypePacket` line 230 in the reported crash, same pattern in `BlockTypeNet`). A concurrent insert into one of those dictionaries throws `InvalidOperationException("Collection was modified")` on the pool thread, and `TyronThreadPool.QueueTask` wraps no try/catch around Action tasks, so the unhandled exception kills the process. We hit this on ~1/15 boots under CI load, with crashes in both the item and block loops.

One honest correction to what I wrote on the issue: stratum.15 queues the build at boot exactly like vanilla; there is no lazy build in the code. What looked lazy in our traces was pool starvation: the global pool is capped at `SetMaxThreads(10, 1)` and several long-running boot tasks occupy workers, so on a loaded 4-core runner the assets build can start late and overlap much more of the boot. The overlap window grows with contention, which is why constrained CI hit the race far more often than dev machines. We could not statically pin the exact mutator (none of the vanilla mod `Start()` paths register classes that late), which is exactly why the fix is structural rather than a lock around one suspect.

**The fix**: call `BuildServerAssetsPacket()` synchronously at the same boot point. Three lines earlier the tag registries are finalized and locked, worldgen startup has not fired, and the server threads have not started: nothing can mutate what the encoders read, whichever mutator used to win the race. It also removes the first join's worst case, a 10s `WaitOnBuildServerAssetsPacket` timeout followed by a synchronous rebuild when the pool-starved build had not finished. `StartBuildServerAssetsPacket` is left in place (now unused) to keep the patch surface minimal against vanilla updates.

Note for upstream: the race exists identically in vanilla (the window is just narrower there); this fix is a good candidate to report to Anego once validated here.

## Type

- [x] Bug fix
- [ ] Performance
- [ ] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

- [x] `.\scripts\extract-patches.ps1` ran clean (Linux: `scripts/extract-patches.sh`; only `ServerMain.cs.patch` changed, offset renumbering aside).
- [x] Tested on a real server start, not just compilation (see below).

## Performance numbers

Boot to "Dedicated Server now running", same indev tree with and without the fix, 3 runs each on a 24-core Linux machine: 13.1 / 13.1 / 14.6s before, 14.1 / 14.6 / 15.1s after. About +1 to +1.5s of serialized build at boot (the build also contains ~170ms of pacing sleeps that only made sense off-thread; left untouched for minimal diff, happy to drop them if wanted). In exchange, the first join never pays the 10s timeout + rebuild path.

## Validation

- Full [StratumParity](https://github.com/Pixnop/StratumParity) differential suite (17 scenarios, real boots, joins, assets sends) green against a release-layout install overlaid with the fixed build: no behavior change detected against vanilla.
- Our CI retries exactly once on this crash signature ("Collection was modified" / "SendServerAssets failed"); once a build with this fix ships in a tag, that retry should never fire again, which gives us a permanent regression sentinel.

## Related issues

Fixes #151.

### DIFF
--- a/patches/VintagestoryLib/Vintagestory.Server/ServerMain.cs.patch
+++ b/patches/VintagestoryLib/Vintagestory.Server/ServerMain.cs.patch
@@ -1,5 +1,5 @@
 diff --git a/VintagestoryLib/Vintagestory.Server/ServerMain.cs b/VintagestoryLib/Vintagestory.Server/ServerMain.cs
-index 1017be9..8d24994 100644
+index 1017be9..a050072 100644
 --- a/VintagestoryLib/Vintagestory.Server/ServerMain.cs
 +++ b/VintagestoryLib/Vintagestory.Server/ServerMain.cs
 @@ -27,10 +27,13 @@ using VintagestoryLib.Server.Systems;
@@ -456,7 +456,29 @@ index 1017be9..8d24994 100644
  		if (progArgs.AddOrigin != null)
  		{
  			foreach (string item in progArgs.AddOrigin)
-@@ -1482,17 +1669,19 @@ public sealed class ServerMain : GameMain, IServerWorldAccessor, IWorldAccessor,
+@@ -1280,11 +1467,20 @@ public sealed class ServerMain : GameMain, IServerWorldAccessor, IWorldAccessor,
+ 		EnterRunPhase(EnumServerRunPhase.WorldReady);
+ 		if (exitState.Exit)
+ 		{
+ 			return;
+ 		}
+-		StartBuildServerAssetsPacket();
++		// Stratum: build the assets packet synchronously instead of on a pool thread
++		// (issue #151). The background build raced late boot work: the block/item
++		// encoders enumerate shared ClassRegistry dictionaries, and a concurrent
++		// insert threw "Collection was modified" on a pool thread with no handler,
++		// killing the process (~1/15 boots under load). Here the registries are
++		// finalized and locked and neither worldgen startup nor the server threads
++		// have run yet, so nothing can mutate what the encoders read. This also
++		// removes the first join's worst case: a 10s WaitOnBuildServerAssetsPacket
++		// timeout followed by a synchronous rebuild.
++		BuildServerAssetsPacket();
+ 		Logger.StoryEvent(Lang.Get("The center unfolding..."));
+ 		Logger.Event("Starting world generators...");
+ 		ModEventManager.TriggerWorldgenStartup();
+ 		if (exitState.Exit)
+ 		{
+@@ -1482,17 +1678,19 @@ public sealed class ServerMain : GameMain, IServerWorldAccessor, IWorldAccessor,
  					ServerSystem serverSystem = Systems[i];
  					long num = elapsedMilliseconds - serverSystem.millisecondsSinceStart;
  					if (num > serverSystem.GetUpdateInterval())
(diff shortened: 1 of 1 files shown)


trevorftp (approved):
Looks good to me!

One tiny thing and for sure not a blocker, this also moves all the `block.FreeRAMServer()` calls to before worldgen startup now instead of whenever the pool thread got around to it. Pretty sure that's fine since it was already racy in vanilla before, just figured I'd flag it to potentially look into later,, send it!

Pixnop:
Good eye on FreeRAMServer. I think the move is strictly safer: before, those calls ran whenever the pool thread got scheduled, which could be mid-worldgen or mid-first-join; now they run at a deterministic point before either starts. Same work, predictable timing. If anything ever does show up around it, the sync call site makes it easy to bisect. Thanks for the fast review, merging.

Issue #147, opened 2026-07-13
StratumParity: external vanilla-parity test suite, adopt or wire into CI


Pixnop:
### Problem

Stratum's promise is server performance without changing vanilla gameplay behavior. Nothing verifies that promise today beyond the boot smoke test, so a patch that shifts a gameplay outcome can land with every check still green.

### Proposed change

Adopt [StratumParity](https://github.com/Pixnop/StratumParity), a differential suite that runs the same scenarios against a vanilla install and a Stratum install (the official release zip, bootstrapped normally) and requires the outcomes to match. It is built on [Atlas](https://github.com/Pixnop/Atlas), a harness that boots a real headless VS server inside the test process, and Atlas scenarios already run in Nimbus, so the harness is exercised elsewhere in this ecosystem. An environment variable selects the install, CI runs the suite on both flavors, and the reports are diffed.

Scenarios come in two families. Parity scenarios assert the same thing on both flavors, so any divergence is a regression. Probe scenarios measure what Stratum changes on purpose, then confirm the config toggle restores vanilla behavior.

Green on `v1.22.3-stratum.14`: server boots and ticks, block placement reads back identically, a joined player survives the packet limiter, the entity distance-band throttle behaves as documented (far entities tick 1 in 10 by default, measured, and full rate again with `EntityTicking.Enabled: false`), and vanilla commands behave identically (`/time add`, unknown-command error codes) while `/stratum` and `/sethome` exist only on Stratum.

Two ways to use it: wire it into CI as a workflow that builds the server and runs the suite on pull requests, or keep it external and treat its weekly run against each stable release as an out-of-tree smoke test. The author is willing to draft the CI workflow.

### Alternatives considered

Keep relying on the boot smoke test alone. That leaves every behavioral guarantee unchecked between releases.

### Scope

- [x] This is a server-side change. It does not require a custom client.

Synthetic clients must be driven to the `Playing` state before the distance throttle counts them (`IsPlayingClient` filter). Atlas reaches `Playing` natively as of 0.9.0.

Zaldaryon:
Thanks for building and sharing this, StratumParity is a real gap this project has. The two behavioral notes are useful too.

Filed a fix for the second one (`stratum-performance.json` only read when `stratum.json` already exists, so a data dir provisioned with just a performance sidecar loses it on first boot): #159. The `IsPlayingClient` note looks like correct, intentional behavior on our side rather than a bug, worth documenting for anyone writing scenarios against the throttle.

[...]

PR #173, opened 2026-07-19
Document entity tick throttling semantics

Pixnop:
## Summary

Documentation for `Performance.EntityTicking`, following up on #147 where Zaldaryon noted the Playing-client anchoring is "worth documenting for anyone writing scenarios against the throttle".

`docs/ENTITY-TICKING.md` covers, verified against the current indev implementation (`ServerSystemEntitySimulation.cs`, `StratumConfig.cs`):

- The four distance bands, their defaults, and the clamping rules.
- The anchoring semantics: players with a spawned entity only (Playing state), per dimension, and the no-anchor case (empty server or wrong dimension) ticking everything at the very-far stride.
- Every exemption in evaluation order: player entities, class toggles, moving entities (with the dropped-item trap), excluded mod domains (with the VSVillage rationale from the code comment).
- The ambient fauna tier, the creature budget and its no-effect-in-parallel-mode caveat, and the vanilla-restore toggle.
- A "testing against the throttle" section with the field-tested traps: Playing-state requirement, motionless probes, geometry derived from actual player positions, exact-count assertions. These are the exact pitfalls the StratumParity suite hit and solved while pinning these semantics down.

## Type

- [ ] Bug fix
- [ ] Performance
- [ ] New feature
- [ ] Refactor or cleanup
- [x] Docs or build

## Checklist

- [x] `.\scripts\extract-patches.ps1` ran clean. (Docs-only change.)
- [x] `dotnet build VintageStory.slnx -c Release` is green. (Not affected.)
- [x] Every vanilla edit has a `// Stratum` marker. (No code touched.)
- [x] Tested on a real server start, not just compilation. (Behavior claims verified against the implementation and exercised by the StratumParity suite.)

## Related issues

Follows up on #147.

### DIFF
--- a/docs/ENTITY-TICKING.md
+++ b/docs/ENTITY-TICKING.md
@@ -1,0 +1,97 @@
+# Entity tick throttling
+
+Stratum throttles server-side entity ticking by distance to the nearest player
+(`Performance.EntityTicking` in `stratum.json` or the `stratum-performance.json`
+sidecar, enabled by default). This page documents the exact semantics: what
+anchors the distance bands, what is exempt, and what to expect when testing
+against the throttle. All defaults below are the shipped values; every knob
+lives under `Performance.EntityTicking`.
+
+## Distance bands
+
+Each entity ticks every Nth server tick based on the 3D euclidean distance to
+the **nearest anchoring player**:
+
+| Band | Distance (blocks) | Interval | Key |
+|---|---|---|---|
+| Near | <= 32 | every tick | `NearDistanceBlocks` |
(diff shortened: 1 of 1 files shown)


trevorftp:
Looks great, maybe we can add this to the wiki instead? Eventually all docs related things will be here anyhow.
https://github.com/StratumServer/StratumDocs

and the wiki removed.

Pixnop:
Makes sense, moved it: StratumServer/StratumDocs#1 carries the same content adapted to the docs repo's format (Configuration section, page 3). I'll delete the docs/entity-ticking branch here once that one lands.

PR #175, opened 2026-07-19
Chunk generation now occurs through the Dispatcher -> Workers system

### LINKED ISSUE #176: Worlds saved with .16-indev 4 can no longer be launched with previous Stratum versions or the Vanilla Server
### Stratum version

.16-indev.4

### Base game version

1.22.3

### OS and .NET runtime

Windows 10.0.26100 Build 26100, .NET 10.0.10

### Steps to reproduce

1. Run Stratum 16 indev-4 (default settings)
2. Close Server
3. Attempt to launch the same world on either a previous version of Stratum or with the Vanilla Vintage Story server (default settings).

### Expected behavior

Clean Launch, normal functions.
I am uncertain if backwards compatibility was intended, indev 4 seems to have broken it though.

### Actual behavior

WorldGen loops for a minute before throwing an error (incl. below). This also prevents normal server functions and clients from connecting.

### Server log

```text
19.7.2026 12:00:56 [Event] Loading 7x7x12 spawn chunks...
19.7.2026 12:00:56 [Event] 0% (49 in queue)
19.7.2026 12:00:58 [Event] 0% (49 in queue)
19.7.2026 12:00:59 [Event] 0% (49 in queue)
19.7.2026 12:01:01 [Event] 0% (49 in queue)
19.7.2026 12:01:02 [Event] 0% (49 in queue)
19.7.2026 12:01:04 [Event] 0% (49 in queue)
19.7.2026 12:01:05 [Event] 0% (49 in queue)
19.7.2026 12:01:07 [Event] 0% (49 in queue)
19.7.2026 12:01:08 [Error] Attempting to force generate chunk columns from 15997,15997 to 16003,16003
19.7.2026 12:01:08 [Error] 0 additional worldgen threads active, number of 'undone' chunks is 49
19.7.2026 12:01:08 [Error] Column 15997,15997 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 15997,15998 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 15997,15999 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 15997,16000 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 15997,16001 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 15997,16002 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 15997,16003 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 15998,15997 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 15998,15998 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 15998,15999 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 15998,16000 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 15998,16001 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 15998,16002 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 15998,16003 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 15999,15997 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 15999,15998 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 15999,15999 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 15999,16000 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 15999,16001 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 15999,16002 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 15999,16003 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16000,15997 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16000,15998 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16000,15999 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16000,16000 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16000,16001 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16000,16002 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16000,16003 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16001,15997 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16001,15998 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16001,15999 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16001,16000 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16001,16001 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16001,16002 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16001,16003 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16002,15997 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16002,15998 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16002,15999 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16002,16000 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16002,16001 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16002,16002 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16002,16003 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16003,15997 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16003,15998 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16003,15999 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16003,16000 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16003,16001 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16003,16002 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Column 16003,16003 has reached pass 7 (in original req)
19.7.2026 12:01:08 [Error] Mod exception during event OnWorldgenStartup. Will skip to next event
19.7.2026 12:01:08 [Error] Exception: Somehow worldgen has become stuck in an endless loop, please report this as a bug!  Additional data in the server-main log
   at Vintagestory.Server.ServerSystemSupplyChunks.loadChunkAreaBlocking(Int32 chunkX1, Int32 chunkZ1, Int32 chunkX2, Int32 chunkZ2, Boolean isStartupLoad, ITreeAttribute chunkGenParams) in VintagestoryLib\Server\Systems\World\LoadThread\SupplyChunks.cs:line 1907
   at Vintagestory.Server.ServerSystemSupplyChunks.InitWorldgenAndSpawnChunks() in VintagestoryLib\Server\Systems\World\LoadThread\SupplyChunks.cs:line 1642
   at Vintagestory.Server.ServerSystemLoadAndSaveGame.OnWorldgenStartup() in VintagestoryLib\Server\Systems\World\LoadThread\LoadSaveGame.cs:line 203
   at Vintagestory.Server.ServerEventManager.<>c.<TriggerWorldgenStartup>b__122_0(Action dele) in VintagestoryLib\Server\ServerEventManager.cs:line 492
   at Vintagestory.Common.EventManager.Trigger[T](Delegate[] delegates, String eventName, Action`1 onDele, Action onException) in VintagestoryLib\Common\EventManager.cs:line 54



19.7.2026 12:18:49 [Notification] Client 2 uid REDACTED attempting identification. Name: REDACTED
19.7.2026 12:18:49 [Error] Exception: Object reference not set to an instance of an object.
   at Vintagestory.Server.ServerMain.GetSpawnPosition(String playerUID, Boolean onlyGlobalDefaultSpawn, Boolean consumeSpawn) in VintagestoryLib\Server\ServerMain.cs:line 1675
   at Vintagestory.Server.ServerMain.FinalizePlayerIdentification(Packet_ClientIdentification packet, ConnectedClient client, String entitlements) in VintagestoryLib\Server\ServerMainNetworking.cs:line 514
   at Vintagestory.Server.ServerMain.PreFinalizePlayerIdentification(Packet_ClientIdentification packet, ConnectedClient client, String entitlements) in VintagestoryLib\Server\ServerMainNetworking.cs:line 341
   at Vintagestory.Server.ServerMain.<>c__DisplayClass332_1.<VerifyPlayerWithAuthServer>b__1() in VintagestoryLib\Server\ServerMainNetworking.cs:line 297
   at Vintagestory.Server.ServerMain.ProcessMainThreadTasks() in VintagestoryLib\Server\ServerMain.cs:line 3099
   at Vintagestory.Server.ServerMain.ProcessMain() in VintagestoryLib\Server\ServerMain.cs:line 950
   at Vintagestory.Server.ServerMain.Process() in VintagestoryLib\Server\ServerMain.cs:line 886
```

### Checks

- [x] I confirmed this does not happen on a stock vanilla server.
- [x] I am running the latest Stratum release.

tehtelev:
## Summary

What does this PR change (compared to vanilla)?
- **Pipeline instead of "1 thread = 1 stage."** In vanilla, CreateAdditionalWorldGenThread(num, num+1, …) rigidly assigned a thread to a single stage, while GenerateChunkColumns_OnSeparateThread polled for "its" work → threads were idle. Now, a separate DispatcherThreadLoop dispatches tasks to the readyTasksQueue (BlockingCollection<DispatchedTask>) channel via ScheduleReadyTasks(), and GeneratorThreadLoop() takes (request, claimedStage) and distributes the chunk through one stages. Stage occupancy is determined by the global stratumStageBusy[] + atomic Interlocked.CompareExchange. Result: N threads dynamically divide the stages of different chunks → almost linear boost (≈2.0–2.18 on 3 threads instead of the vanilla 1.1–1.3).
- **Event-driven waiting, 0% CPU idle**. Instead of Thread.Sleep(1) + polling, we use a parking queue on dispatcherWake (AutoResetEvent) and a blocking TryTake. Any state change wakes the dispatcher via SignalDispatcher(). High CPU in tests isn't due to "gluttony," but to previously sleeping cores that are now being counted.
- **Player prioritization.** TryBuildStratumPriorityList() builds a list once per tick based on the distance to the nearest player (StratumMinSqDistToPlayers) with position prediction based on the motion vector (pos + Motion * PredictionSeconds * 30). Top-K — bounded by the stratumTopKHeap heap in O(N log K), the result is cached per tick (stratumPriorityListCache), and when exhausted, graceful fallback to FIFO. Chunks in front of the player are generated first.
- **Parallel column reading from the database.** TryLoadChunkColumn received a path via StratumChunkReadPool + Parallel.For (MaxDegreeOfParallelism = min(WorkerCount, chunkMapSizeY)) — Y levels are read simultaneously; enabled by configuration and only when advantageous.
- **Fewer allocations per tick.** Buffers are moved to fields and reused via .Clear(): stratumMoveBuffer, stratumCandidatesBuffer, stratumDeleteChunksList/MapChunksList/RegionsSet, player arrays stratumPlayerCx/Cz. In testChunkExistsQueue, the lambda closure has been replaced with val.Exists = exists; EnqueueMainThreadTask(val.Complete) — without closure allocation. Less garbage → shorter GC pause.
- **O(N M) → O(N + M).** In TryRestoreGeneratedStructures, the nested .Any(...) has been replaced with HashSet<Vec3i> + Contains.
- **Multithreading correctness:**
Throttling of feeds in loadChunkAreaBlocking — a loop limiting free capacity to prevent a large area from overflowing the queue and causing an emergency EnsureQueueSpace call, pausing all threads. Pause deadlock fix: after a timeout, PauseAllWorldgenThreads now does Interlocked.Exchange(ref pauseAllWorldgenThreads, 0) + SignalDispatcher() (in vanilla, the flag wasn't cleared → potential indefinite hang).
Guaranteed stage release via handedOff/holdingStage flags in try/finally — a "leaked" busy stage no longer blocks the pipeline.
Non-blocking dispatcher — Monitor.TryEnter(dispatcherLock) + InvalidOperationException handling when modifying the queue.
Stage boundaries: the chunk thread only takes None/Done, all worker stages are taken by workers; in ensurePrettyNeighbourhood, the threshold has been raised to Terrain2 (early start without waiting for neighbors).
- **Configurability and clean shutdown.** Prioritization and parallel IO are controlled by StratumRuntime.Config and degrade gracefully to vanilla FIFO when !Enabled is selected. OnSeperateThreadShutDown/Dispose set dispatcherShuttingDown, wake up the dispatcher, and Join(2000) and readyTasksQueue.CompleteAdding() deterministically exit workers.

**Other changes:**
- ChunkServerThread: Avoiding Func<> + DisplayClass allocation for each call. In the vast majority of cases, this entry already exists in the dictionary.
- ServerMapChunk: Implements compatibility between extended Stratum chunk generation stages and vanilla stages.
- ServerSystemUnloadChunks: Added protection for the chunk unload method call from being called concurrently from different threads at the same time. This could potentially cause saving issues.
- StratumConfig: Added automatic trimming of the "MaxColumnsPerSecond", "MaxPendingColumnQueue", and "MaxWorkerColumnQueue" values ​​to prevent queue overflow. If this does occur, you should manually increase the RequestChunkColumnsQueueSize value. PauseBelowTps decreased from 20 to 15.0. StratumChunkPriorityConfig parameter "Enabled" is now true by default.
- StratumPregenManager: Added automatic TTL reduction function to speed up unloading of completed chunks from memory, now the tick occurs not every 1000 ms, but every 50 ms, since the idle phases began to take up a significant amount of time

## Type

- [ ] Bug fix
- [x] Performance
- [ ] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

## Performance numbers

Pregeneration performance testing summary tables.

image
image

Test computer:
Intel I7-11700K 4.6 GHz,
64 GB RAM 3200 MHz

Flight test on 6 generation threads

Below are more details on each test configuration.
Tested pre-generation of 10200 chunk columns on the same seed.

<details>
  <summary>Vanilla 1.22.3</summary>
Command for generating chunks

/wgen pregen 50

Configuration file settings:
- servermagicnumbers.json
"UncompressedChunkTTL": 3000,
"CompressedChunkTTL": 9000,

image

**2 threads (without mods)**
image

**6 threads (without mods)**
image

**Memory allocations  (3 threads)**
image

</details>

<details>
  <summary>Stratum 15</summary>
Command for generating chunks

/stratum pregen start rect 15950 15950 16050 16050

Configuration file settings:
- servermagicnumbers.json
"UncompressedChunkTTL": 3000,
"CompressedChunkTTL": 9000,
- stratum-performance.json
"Pregen": {
    "Enabled": true,
    "MaxColumnsPerSecond": 500,
    "MaxScansPerSecond": 4096,
    "MaxPendingColumnQueue": 500,
    "MaxWorkerColumnQueue": 1000,
    "MaxLoadedChunkColumns": 4096,
    "MaxRadiusChunks": 1000,
    "MaxAreaColumns": 1000000,
    "PauseWhenPlayersOnline": true,
    "PauseBelowTps": 15.0,
    "ProgressLogIntervalSeconds": 60
  },

This version essentially contains the vanilla generator, but with its own pre-generation function.
At the same time, numerous optimization patches are already included, which reduce the generation time of each generation stage and memory allocations.

image

**2 threads (without mods)**
image

**6 threads (without mods)**
image

**Memory allocations (3 threads)**
image

</details>

<details>
  <summary>Now</summary>
Command for generating chunks

/stratum pregen start rect 15950 15950 16050 16050

Configuration file settings:
- servermagicnumbers.json
"UncompressedChunkTTL": 15000,
"CompressedChunkTTL": 45000,
When pregeneration is enabled, these values ​​are reduced by 5 times.

- stratum-performance.json
"Pregen": {
    "Enabled": true,
    "MaxColumnsPerSecond": 500,
    "MaxScansPerSecond": 4096,
    "MaxPendingColumnQueue": 500,
    "MaxWorkerColumnQueue": 1000,
    "MaxLoadedChunkColumns": 4096,
    "MaxRadiusChunks": 1000,
    "MaxAreaColumns": 1000000,
    "PauseWhenPlayersOnline": true,
    "PauseBelowTps": 15.0,
    "ProgressLogIntervalSeconds": 60
  },

It can be seen that the optimal value of threads for generation is **3-4**. Further increase in threads increases the load on the CPU without a noticeable increase in performance.

image

**Examples of load distribution across threads.**

**2 threads (without mods)**
image

**6 threads (without mods)**
image

**Memory allocations  (3 threads)**
image

</details>

## Related issues

Closes #176


Pixnop (commented):
Saw this went to draft while I was mid-review, so here is what I found anyway: hope it saves you a round-trip while you iterate.

First: I like where this is going. Moving generation to a dispatcher/workers pipeline is the right shape, and I wanted to give it a real review, so I built this branch and ran the full StratumParity differential suite against it: 17/17 green (chunk persistence round-trips, random ticks, entity ticking all behave). The behavioral surface looks solid. The concurrency layer has a few real problems though, all fixable. Line numbers below are from the bootstrapped working tree of this branch.

## Blocking before ready-for-review, I think

**1. The claimed stage does not travel with the request (ServerSystemSupplyChunks.cs, dispatcher ~236-261, worker ~326-327).** The dispatcher reads `stage = req.CurrentIncompletePass_AsInt`, CAS-claims `stratumStageBusy[stage]`, then enqueues the bare request. The consumer re-derives `currentStage = request.CurrentIncompletePass_AsInt` and the comment asserts "Dispatcher already claimed the stage". That assertion can be false: if the dispatcher's read happens while a worker is finishing pass S on that same column (the worker releases S at ~369/~383 after `currentpass` moved to S+1), the dispatcher claims a stale S and the consumer derives S+1. From there: busy[S] is claimed by nobody-in-particular and never released, so pass S wedges permanently for the whole server (every column needing S stalls, `loadChunkAreaBlocking` eventually throws "worldgen stuck"); the consumer runs pass S+1 without holding its claim, so a second thread can legitimately claim S+1 for another column and you get two threads inside the same non-reentrant pass, the exact thing `stratumStageBusy` exists to prevent; and the consumer's finally releases busy[S+1], which may yank it from a legitimate holder. Suggested fix: enqueue `(request, claimedStage)` tuples, have the consumer use and release the stage it received, and have the dispatcher re-verify `CurrentIncompletePass_AsInt == stage` after the CAS (release and skip on mismatch).

**2. Claimed stage leaks on any exception between CAS and Add (dispatcher ~243-261).** `ensurePrettyNeighbourhood` runs after the claim and can throw (its `EnsureQueueSpace` path reaches `UnloadGeneratingChunkColumns`, which contains a literal `throw new Exception("Chunkrequest no longer in queue? Race condition?")`, plus whatever the DB layer throws), and `readyTasksQueue.Add` throws `InvalidOperationException` after `CompleteAdding`. Both paths bypass any release: the IOE lands in the catch at ~265 that is documented as the "collection was modified during iteration" case (two unrelated failure modes share one silent handler), everything else lands in the outer catch at ~272. Either way the claimed stage is never freed: same permanent wedge as above. The claim needs a try/finally of its own, and the Add-after-shutdown case deserves its own handling rather than riding the iteration-race catch.

**3. Same leak shape in the worker pipeline (worker ~356-383).** The worker CAS-claims `nextStage`, then calls `ensurePrettyNeighbourhood` before transferring `currentStage = nextStage`. If that call throws, the outer catch fires and the finally releases only `currentStage`: the claimed `nextStage` leaks permanently. Releasing `nextStage` on the exception path (or transferring before the risky call and releasing the old stage) closes it.

**4. `reusableStream` is now shared across threads without a common lock (ServerSystemUnloadChunks.cs:37, :215, :429).** The field is static. `UnloadGeneratingChunkColumns` uses it under the new `unloadGenLock` (:387/:429), which is correct in isolation, but the chunk-thread save path at :215 uses the same stream with no lock at all. The PR itself notes "saving can now be called from different threads" and added dedicated `stratumUnloadGenDbChunkBuf1/2` lists for exactly this reason; the stream needs the same treatment. Two threads serializing through one `FastMemoryStream` interleave bytes into the blobs written to the save DB: silent world corruption, and it triggers precisely when the server is overloaded and both paths have dirty chunks to flush. A dedicated stream for `UnloadGeneratingChunkColumns` (like the buffers) is the minimal fix.

## Likely, please double-check

- `loadChunkAreaBlocking`'s new wait loop (`while (Capacity - Count < 200 ...) { CleanupRequestsQueue(); ScheduleReadyTasks(); Sleep(1); }`) can only drain stages 1..5. Requests at stage 0 (load) and Done (finalize) are processed by the chunk thread, which is the thread spinning in this loop: a backlog dominated by those stages never drains and the loop never exits.
- `FullyClearGeneratingQueue` clears `requestedChunkColumns` without flagging anything Disposed, and the worker never rechecks `request.Disposed` after `TryTake`: entries already in `readyTasksQueue` survive the wipe and get generated into columns that were evicted from every registry (never delivered, never saved). A Disposed recheck after TryTake (releasing the claimed stage) plus draining `readyTasksQueue` on the wipe would cover it.
- The global TTL boost does not look restored on the pressure-pause path (low TPS / full queue), and the `pauseAllWorldgenThreads` reset-on-timeout can race the workers' increment. Both worth a second look.

## Intentional? Then say so

Two defaults change silently in a PR labeled performance: `StratumChunkPriorityConfig.Enabled` false to true, and `PauseBelowTps` 20.0 to 15.0. Both may well be right, but they change behavior for every existing server and belong in the summary (arguably in their own commit).

## Housekeeping

- `StratumConfig.cs` was fully reindented tabs to spaces, which turns a real ~60-line change into a 2900-line diff and hides the two default flips above. The repo convention is tabs (CONTRIBUTING); re-tabbing would shrink this PR by two thirds. Same spaces crept into vanilla code in the patches.
- The template's Summary and Performance numbers sections are empty; CONTRIBUTING asks for before/after numbers on hot paths, and this is the hottest path there is. Even one "time to generate N chunks around a moving player" comparison would do.
- A few leftovers: Russian comments (`// было stratumDbChunkBuf1`, ServerSystemUnloadChunks.cs:425-427), some pre-existing `// Stratum:` markers dropped while the vanilla edit remains, unused `System.Threading.Channels`/`System.Diagnostics` usings.

Happy to re-run the full suite plus a boot-loop on the next revision, and to pair on the claim-transfer fix if useful: the tuple change is mechanical but touches every enqueue site. The 17/17 above says the happy path is already right; these fixes are about keeping it right under load and shutdown.

tehtelev:
Regarding the bugs you mentioned.

1. Claimed stage does not travel with the request
Confirmed, fixed. The dispatcher now enqueues a (request, claimedStage) tuple (DispatchedTask struct), the consumer uses the received stage instead of re-deriving it, and the dispatcher re-verifies CurrentIncompletePass_AsInt == stage immediately after the CAS (releases and skips on mismatch).
2. Claimed stage leaks on exception between CAS and Add
Confirmed, fixed. The section between CAS and Add is now wrapped in its own try/finally with a handedOff flag — any exception (from ensurePrettyNeighbourhood, Add-after-CompleteAdding, or anything else) releases the stage in finally. The single catch (InvalidOperationException) is split into two filtered catches to distinguish collection-modification-during-iteration from channel-completed.
3. Same leak shape in the worker pipeline
Confirmed, fixed. Ownership transfer (currentStage = nextStage) now happens before the risky ensurePrettyNeighbourhood call, and the previous stage is released immediately. If the call throws, the finally block releases currentStage which already equals nextStage — no leak.
4. reusableStream shared across threads
Not an issue. The field is marked [ThreadStatic] — each thread (chunk-thread, dispatcher, each worker) gets its own FastMemoryStream instance. There is no shared stream and no interleaving path. The unloadGenLock exists to protect shared collections (requestedChunkColumns, loadedMapChunks) and DB writes from concurrent mutation, not the stream.

Everything that is indicated for double-checking is ok. That's how it was intended.

Other changes:
- StratumConfig.cs and StratumPregenManager.cs indents have been changed to tabs. Subsequent files will now also have tab indents;
- Fixed the bug mentioned [here](https://github.com/StratumServer/Stratum/issues/176). Saves now have built-in backward compatibility with regular saves. Saves made on 16.indev 4 will unfortunately remain broken. But this is a test branch, so backups are everything.

Pixnop (commented):
Re-ran everything on the updated branch. Short version: this is in great shape now.

**Points 1-3 are fixed, and cleanly.** The `DispatchedTask(request, claimedStage)` tuple with the post-CAS re-verify and the `handedOff` try/finally closes the claim-travel race exactly; the worker's ownership transfer before `ensurePrettyNeighbourhood` closes the nextStage leak; and you even split the conflated `InvalidOperationException` catch. Verified by reading the regenerated sources, not just the diff.

**Point 4 retracted: you are right and I was wrong.** `[ThreadStatic]` has been on `reusableStream` since the vanilla decompile: my grep showed the field line without the attribute line above it, so my review claimed a cross-thread race that never existed. No shared stream, no corruption path. Sorry for the noise.

**Empirical re-validation:**
- Full StratumParity suite against this branch: 17/17, no behavioral divergence.
- Cross-version round-trips: a world created on this branch reloads clean on stratum.15 (the exact downgrade that bit the official server), and a stratum.15 world upgrades clean onto this branch. The `ToPersistedPass`/`FromPersistedPass` boundary does its job.

**On indev.4-written saves:** your comment and #176 already answer what I was going to raise (their on-disk values are internal, so the translation shifts frontier chunks a pass forward): accepting the breakage on a test branch with backups is a reasonable call. Just noting the official server's current save is in that state, for whenever it migrates.

**Smaller leftovers, non-blocking:** `FullyClearGeneratingQueue` still wipes `requestedChunkColumns` without draining `readyTasksQueue` (harmless-er now: the worker's pass recheck releases the claim properly, but leftover tasks still generate into evicted columns), and the `loadChunkAreaBlocking` wait loop still can't drain a backlog dominated by stage 0/Done (chunk-thread-only stages, and that thread is the one spinning). You said the double-check items are as intended: works for me. The PR body's Summary and Performance numbers are still empty though: with the vanilla cross-chunk bug you uncovered, this PR has a better story to tell than most.

Side note so nobody bisects chunk code for it: reloading a Stratum data dir on pure vanilla fails much earlier on the roles split (`default group code suplayer` fatal): unrelated to this PR.

trevorftp (changes requested):
The earlier stage ownership fixes look great, but I found a few things that may need fixed before we push this fully. I would love for @Pixnop to review and double check my findings and the code as well.

### 1. Reduced thread mode has no generation worker.
https://github.com/StratumServer/Stratum/blob/8ef99056b4d6ec9b9e10a208b9edcac106e0d312/patches/VintagestoryLib/Vintagestory.Server/ChunkServerThread.cs.patch#L28-L31 sets `additionalWorldGenThreadsCount` to 0 when `ReducedServerThreads` is enabled.


https://github.com/StratumServer/Stratum/blob/8ef99056b4d6ec9b9e10a208b9edcac106e0d312/patches/VintagestoryLib/Vintagestory.Server/ServerSystemSupplyChunks.cs.patch#L159-L163 then creates zero generation workers.

The dispatcher places stages 1 through `Done - 1` into `readyTasksQueue` at
https://github.com/StratumServer/Stratum/blob/8ef99056b4d6ec9b9e10a208b9edcac106e0d312/patches/VintagestoryLib/Vintagestory.Server/ServerSystemSupplyChunks.cs.patch#L253-L282 while the chunk thread only handles `None` and `Done` at https://github.com/StratumServer/Stratum/blob/8ef99056b4d6ec9b9e10a208b9edcac106e0d312/patches/VintagestoryLib/Vintagestory.Server/ServerSystemSupplyChunks.cs.patch#L826-L840 With no workers, a request advances to stage 1 and then has nowhere to run.

We need either a zero-worker fallback where the chunk thread runs every stage, or a guarantee that at least one actual generation worker exists.


### 2. The new pregen budget does not honor low configured rates.

The accumulator calculates the correct fractional budget at https://github.com/StratumServer/Stratum/blob/8ef99056b4d6ec9b9e10a208b9edcac106e0d312/sources/VintagestoryLib/Vintagestory.Server/StratumPregenManager.cs#L215-L225 however https://github.com/StratumServer/Stratum/blob/8ef99056b4d6ec9b9e10a208b9edcac106e0d312/sources/VintagestoryLib/Vintagestory.Server/StratumPregenManager.cs#L229 forces at least one column on every 50 ms invocation.
`MaxColumnsPerSecond = 1` can therefore run close to 20 columns per second instead of one. Ticks without a full column of accumulated credit should do no work.


### 3. The pause timeout race is still present.
The timeout resets the shared counter to 0 at https://github.com/StratumServer/Stratum/blob/8ef99056b4d6ec9b9e10a208b9edcac106e0d312/patches/VintagestoryLib/Vintagestory.Server/ServerSystemSupplyChunks.cs.patch#L2363-L2374 A worker can already have passed the positive-value check at https://github.com/StratumServer/Stratum/blob/8ef99056b4d6ec9b9e10a208b9edcac106e0d312/patches/VintagestoryLib/Vintagestory.Server/ServerSystemSupplyChunks.cs.patch#L2394-L2401 then increment the value back to 1 after the reset.

`EnsureQueueSpace` returns when pausing fails and does not call `ResumeAllWorldgenThreads`, so that late acknowledgement can leave world generation paused. The pause request and acknowledgement count should use separate state, or the worker needs to recheck the request and undo its acknowledgement if the pause was cancelled.


### 4. Game version changed... Again...

New version is 1.22.5, so a bootstrap will fail. Thanks Anego :P

Pixnop (commented):
@trevorftp asked me to double-check his findings: I verified all four against the code, and found two more issues in the process. Summary: 1, 2 and 3 confirmed, 4 confirmed and it splits into a repo-wide part and a PR-specific part, plus a dropped fix that I think is the most urgent item here.

## Verification of trevorftp's findings

**1. Reduced thread mode has no generation worker: CONFIRMED.** `additionalWorldGenThreadsCount = 0` under `ReducedServerThreads` (vanilla behaviour, unchanged), the worker-creation loop then runs zero times despite its comment saying "always >0", and the chunk thread's filter is explicit: `if (pass == None || pass >= Done)`. So with `--reducedthreads`, anything at stages 1 through Done-1 has no executor at all. Worth noting the flag is real and supported (Stratum's own PhysicsManager patch honours it), and low-core VPS hosts are exactly the audience.

**2. Pregen budget floor: CONFIRMED, with the exact factor.** `Math.Max(1, maxColumnsThisTick)` guarantees one column per invocation, and the new `ServerSystemStratumPregen` ticks at 50 ms, so `MaxColumnsPerSecond = 1` yields up to 20 columns/s, a 20x overshoot. This is a regression from the last commit specifically: at the old 1000 ms tick the floor was harmless (1 column/s for a 1/s setting); moving to 50 ms turned it into a rate violation.

**3. Pause timeout race: CONFIRMED, and worse than described.** Beyond the check-then-increment window, `PauseWorldgenThreadIfRequired` never decrements its acknowledgement: it breaks out of the wait after 5 s and leaves the counter at 1 or more. With `EnsureQueueSpace` returning on the failure path without calling `ResumeAllWorldgenThreads`, nothing ever resets it, and the dispatcher's `if (pauseAllWorldgenThreads > 0) return false` then refuses to dispatch forever. So the outcome is a permanent worldgen wedge, not a transient stall.

**4. Bootstrap: you are right about 1.22.5, and I measured exactly how much of it is yours.** I bootstrapped four combinations to separate the two problems:

| Baseline | plain indev | this PR |
|---|---|---|
| 1.22.4 (pinned here) | 1 patch fails (`IPlayer`, pre-existing) | **3 fail**, incl. `ServerMain` and `ServerSystemSupplyChunks` |
| 1.22.5 (the new release) | 7 fail | the same 7 fail |

So the 1.22.5 breakage is repo-wide, identical on indev and on this branch: that is a normal version-bump chore for the whole project, not something to hold this PR for. What is specific to this PR is the 1.22.4 column: two extra failures on its own pinned baseline, both core files.

Good news on that part: I checked whether the game update is to blame, and it is, but only cosmetically. Diffing the decompiled vanilla across 1.22.3, 1.22.4 and 1.22.5, those two files change **only in version-string literals** (two lines in `ServerSystemSupplyChunks.cs`, three in `ServerMain.cs`: the disconnect message, the startup log, the `GameVersion` field). Zero functional change. Your hunks fail only because those literals sit inside their context windows, so this is a re-extract on a 1.22.4 tree, not a conflict resolution: nothing of yours actually collides with the update.

(I also tried whether fuzzy application would absorb this class of breakage automatically, since it is going to recur on every patch release: it does not. On 1.22.5, 6 of the 7 still fail under `patch -F5`, because those particular patches modify version-carrying lines themselves. So the bump really does need a re-extract each time.)

**A tooling note that explains how this slipped:** `scripts/bootstrap.sh` prints "N patch(es) failed to apply" but still exits 0 and prints "Bootstrap complete". A build after a failed bootstrap therefore compiles vanilla files for the failed patches silently: on this branch that means none of the dispatcher work is in the binary you just built. Probably worth making the script exit non-zero, happy to send that as a separate PR.

## One more, and I think it is the blocker

**The #151 fix is gone from this branch.** The PR's patch set contains zero occurrences of `BuildServerAssetsPacket` anywhere, while indev's `ServerMain.cs.patch` carries the synchronous call from #172. So merging this branch as-is would silently reintroduce the process-killing assets race we closed on the 19th.

I suspect this is the tooling note above biting rather than a bad merge resolution, and if so it is worth everyone knowing: `ServerMain.cs.patch` is one of the patches that fails to apply on 1.22.4, bootstrap does not fail loudly when that happens, and `extract-patches` then regenerates the patch from a working tree where the failed hunks were never applied. That silently deletes whoever else's work was in that file. Same shape as `ServerSystemSupplyChunks`, which is why I would re-extract from a tree you have verified applied cleanly, then check explicitly: `grep -c "BuildServerAssetsPacket();" patches/VintagestoryLib/Vintagestory.Server/ServerMain.cs.patch` should return 2.

## Empirical validation, and what it revealed about the patch set

Trying to actually run the StratumParity suite against this branch turned out to be the most informative part of this review, because it surfaced one more thing: **no single baseline applies the whole patch set.** On 1.22.4 you lose the dispatcher work (the 3 failures above). On 1.22.3 the dispatcher patches apply, but 5 others fail instead (`AssemblyInfo`, `ServerProgram`, `ServerProgramArgs`, `ServerSystemBlockSimulation`, plus the pre-existing `IPlayer`), so you lose the version bump and the block simulation changes. The mix is what you would expect after the merge: core patches are 1.22.3-era, version-bump patches are 1.22.4-era, and nothing was re-extracted afterwards.

**The design itself is fine, though.** I forced a complete build anyway (1.22.3 tree, applying the stragglers with `patch -F5`), confirmed the dispatcher code was really in the binary (`DispatchedTask` and `ServerSystemStratumPregen` both present), and ran the full suite: **17/17, no behavioral divergence against vanilla**, including chunk persistence round-trips, entity ticking, block tick listeners and random ticks under the new dispatcher.

One note in case it comes up: my first attempt scored 16/17, with `RandomTickProbes.FarPlatform_Should_ConvertOnVanillaAndStayUntouchedOnStratum` reporting 19/1024 blocks converted. That was my build, not your code: `ServerSystemBlockSimulation.cs.patch` was among the ones that had not applied, and that patch is exactly what clamps `blockTickChunkRange` for `LimitRandomTicks`, so the binary had vanilla random-tick range. Applying it made both random-tick scenarios pass. Mentioning it so nobody chases it as a regression later.

So: the architecture behaves and the branch state is what needs fixing. The stage-ownership work from the last round holds up on re-read too, and the dispatcher/workers direction is still clearly the right one. Once the patches are re-extracted on a single tree I will re-run the suite plus the cross-version round-trips on the real baseline.

PR #158, opened 2026-07-14
Fix physics worker thread kept alive

trevorftp:
## Summary

Fixes https://github.com/StratumServer/Stratum/issues/146
Just changed `IsBackground` to `true`.

## Type

- [x] Bug fix
- [ ] Performance
- [ ] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

- [x] `.\scripts\extract-patches.ps1` ran clean.
- [x] `dotnet build VintageStory.slnx -c Release` is green.
- [x] Every vanilla edit has a `// Stratum` marker.
- [x] No vanilla source committed.
- [x] Tested on a real server start, not just compilation.

PR #177, opened 2026-07-23
Fix indev's broken bootstrap, harden the chunk dispatcher, restore mod compatibility

Zaldaryon:
## Summary

This closes out the concurrency and correctness work #175 opened, and finishes a version bump Tsu already started. Eight commits, each with its own bootstrap and build verification, each fixing one problem found by reading the merged dispatcher code and running it.

**The one that matters most first: the patch set on indev does not apply on its own pinned 1.22.4 baseline.** `ServerMain.cs.patch` and `ServerSystemSupplyChunks.cs.patch` fail (they carry 1.22.3-era context around the version-literal lines), bootstrap prints the failure and still exits 0, and a build after that compiles the vanilla files for the two patches that failed, with no error anywhere in the chain. The #172 fix (synchronous `BuildServerAssetsPacket`) is gone from `ServerMain.cs.patch` as a direct result: nobody re-extracted it after this dropped it. Anyone building indev today gets vanilla `ServerMain` and vanilla chunk generation. First commit fixes both patches and makes bootstrap exit non-zero on any failure instead of hiding it.

The rest: a chunk read-pool path that turned every I/O error into "row missing" and regenerated terrain over saved builds as a result, a pause-protocol race that could wedge worldgen for good under one specific interleaving, a dispatcher that scanned the full request queue up to 500 times a second even with nothing to dispatch, the priority-by-player-distance list that never reached the worker stages it was built for, and the big one: #175 renumbered the public `EnumWorldGenPass` enum to fit the Terrain split, which breaks any compiled mod that references the enum. C# inlines enum constants at compile time, so a mod built against vanilla VSAPI that registers at `Vegetation` runs at `TerrainFeatures` instead once this server relabels the numbers, and nothing flags the mismatch. The last commit adds real per-stage timing and reports what it found: `Terrain` alone costs more than the next two stages combined, so the original two-way split is not the equal-load split its own PR description claims, and the next parallelism step needs a different technique than the one this PR uses.

One more thing this branch closes out: Tsu's last commit on indev before this branch (d3ed574) bumped `StratumInfo.BaseGameVersion` to 1.22.5, with a comment on that line asking for `forks.json` and the decompile target to move together. `forks.json` and `Makefile` stayed at 1.22.4, so indev has run in a half-bumped state since that commit landed. Commit 1 in this PR targets 1.22.4 on purpose, matching what was pinned at the time. The last commit finishes the move Tsu started: `forks.json`, `Makefile`, and both bootstrap scripts go to 1.22.5, the whole patch set gets re-extracted against that baseline, and the same verification runs again there.

Each commit message carries the specific reasoning, the interleaving each race needs, and the numbers behind each claim. Read them for the detail, they stand on their own.

## What I tested

`bash scripts/bootstrap.sh` after every commit: 159 patches, zero failures, against whichever baseline that commit targets (1.22.4 through commit 7, 1.22.5 from the last commit on). `dotnet build VintageStory.slnx -c Release`: zero errors after every commit.

Past compilation: built the decompiled server binary and ran it headless several times, direct, not through the launcher, generating real chunk columns end to end, on both baselines. Every run produced zero worldgen exceptions, real dungeons and structures, a clean shutdown. I ran this on the enum-renumbering commit in particular, since a wrong pass number there breaks worldgen order without throwing, again on the timing commit to confirm the added instrumentation leaves generation untouched, and again after the 1.22.5 move to confirm the whole branch reproduces on the new baseline the same way.

I benchmarked throughput before and after: same machine, `/stratum pregen start radius 25` (1961 columns), 6 worker threads forced through `servermagicnumbers.json`, two runs on each side to see if the numbers held.

| Build | Run 1 | Run 2 | Avg |
|---|---|---|---|
| indev, commit 1 alone (patch fix) | 46s | 42s | 44s |
| Commits 1 through 7 (before the 1.22.5 move) | 45s | 43s | 44s |

Ran on commits 1 through 7, before the last commit moved the baseline to 1.22.5. Same throughput both sides. That's the result I expected, not a null one: commits 2, 3, and 6 fix correctness and compatibility bugs with no throughput target, commit 5 does nothing without connected players (pregen has none, by design), and commit 4 targets the idle-dispatcher path, a state a saturated pregen run spends almost no time in since the queue stays fed the whole time. The last commit changes the decompile target, not the dispatcher or chunk-gen code, and the two baselines produce identical diffs for every file this branch edits with dispatcher or chunk-gen logic, so I did not expect and did not re-run the throughput comparison after it.

`make smoke` did not finish in my environment, on either baseline. On 1.22.4 it hit the `StratumInfo.BaseGameVersion`/`forks.json` gap I mentioned above, and the last commit closes that gap. On 1.22.5, after that fix, the launcher's own archive extraction reaches close to the full unpacked size on disk and then the process ends with no exception, no log line, nothing in dmesg. I could not pin down a cause in the time I had, and I cannot rule out something specific to my sandboxed test environment (this launcher path downloads and unpacks a fresh archive itself, outside the bootstrap scripts this PR touches). Every actual code change in this branch goes through the direct headless boot path described above instead, on both baselines and across multiple runs, with real chunk generation and zero exceptions. Flagging the launcher stall here so it does not read as silence: I looked, I could not close it out, and I would rather say so than claim a check I did not finish.

## Type

- [x] Bug fix
- [x] Performance
- [ ] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

- [x] `.\scripts\extract-patches.ps1` ran clean. (`extract-patches.sh` on Linux here, same result: 159 patches, zero failures against the pinned baseline.)
- [x] Tested on a real server start, not just compilation. "What I tested" above covers the full scope of that.

## Performance numbers

See the pregen table above. Throughput holds steady at 44s both sides, 1961 columns, 6 threads: the fixes target correctness and compatibility, not speed, and the one efficiency change in this set (commit 4) targets a workload shape, an idle dispatcher between bursts, that this benchmark does not exercise. The per-stage timing breakdown the last commit adds is real data from that same run:

```
Terrain=46.7% (2721 cols, 14.259ms/col)
TerrainLate=20.2% (2721 cols, 6.162ms/col)
TerrainFeatures=18.5% (2501 cols, 6.149ms/col)
Vegetation=11.6% (2289 cols, 4.196ms/col)
NeighbourSunLightFlood=2.9% (1895 cols, 1.252ms/col)
PreDone=0.2% (1471 cols, 0.131ms/col)
```

## Related issues

None filed yet. The `StratumInfo`/`forks.json` version gap is closed by this PR's last commit, no issue needed there. I can open one for the launcher's silent stall on 1.22.5 archive extraction if that helps, since I could not close that one out myself.


patches/VintagestoryLib/Vintagestory.Client/ClientProgram.cs.patch
tehtelev, line 1: It seems to me that this patch is unnecessary in Stratum?

patches/VintagestoryLib/Vintagestory.Client/GuiScreenDownloadMods.cs.patch
tehtelev, line 1: It seems to me that this patch is unnecessary in Stratum?

patches/VintagestoryLib/Vintagestory.Client/GuiScreenServerDashboard.cs.patch
tehtelev, line 1: It seems to me that this patch is unnecessary in Stratum?

patches/VintagestoryLib/Vintagestory.Common/LoadBalancer.cs.patch
tehtelev, line 1: It seems to me that this patch is unnecessary in Stratum?

tehtelev (changes requested):
Overall, I like it. Visually, it's ok. We'll test it and see if it launches as expected.
Are there just a few patches from the Client part that might have migrated from your Optimum project? I think they should be removed altogether.

patches/VintagestoryLib/Vintagestory.Client/ClientProgram.cs.patch
Zaldaryon: Traced it before answering: Garward added this in 85c3af7 ("Add Linux patch workflow and client lib surface"), the commit that introduced scripts/bootstrap.sh and extract-patches.sh in the first place. It has nothing to do with client behavior. Two fixes: a GLFW ErrorCallback type alias the decompiler leaves ambiguous, and a Vsync cast the decompiled IL emits as an implicit int-to-enum conversion that does not compile. Both are needed for VintagestoryLib.csproj to build at all, since Vintagestory.Client shares the same assembly as the server code. Left it in place.

patches/VintagestoryLib/Vintagestory.Client/GuiScreenDownloadMods.cs.patch
Zaldaryon: Traced it: Tsu added this in 95e6513, "Default ClientModPolicy off, always allow core client mods, repair build pipeline." The C# regex source generator names its generated class after a hash of the pattern plus build context, and that hash changes between ilspycmd runs, so the decompiled call site stops resolving. The fix drops the generator attributes for a plain new Regex(...) call, which sidesteps the instability instead of chasing the hash each time.

I hit the same failure re-extracting against 1.22.5 in this branch's last commit, before I had read this file's history, and reached for the same fix on my own. Left it as it stood.

patches/VintagestoryLib/Vintagestory.Client/GuiScreenServerDashboard.cs.patch
Zaldaryon: Same source as the ClientProgram one: Garward's 85c3af7. JToken.op_Implicit is the compiler's internal name for an implicit-conversion operator; the decompiler emits the internal name as a direct call, which is not legal C# to write by hand, so it does not compile. JToken.FromObject is the public method that does the same conversion. Needed to build the shared assembly, not a client feature change. Left it in place.

patches/VintagestoryLib/Vintagestory.Common/LoadBalancer.cs.patch
Zaldaryon: This one is not client code at all: Tsu wrote it in b50f480, PR #158, "Fix physics worker thread kept alive." One line, thread.IsBackground false to true, so a physics worker thread stops keeping the process alive after an embedded game thread gets abandoned. Lives in Vintagestory.Common. My branch's re-extraction touched nothing here but the diff's line numbers, zero content change. Left it as Tsu wrote it.

Zaldaryon:
Checked git history on every flagged file before answering: none of them came from this branch. `ClientProgram.cs.patch`, `GuiScreenDownloadMods.cs.patch`, and `GuiScreenServerDashboard.cs.patch` (the other two you commented on, `ClientPlatformWindows.cs.patch` and `GuiElementModCell.cs.patch`, look like they got deleted before I could reply) trace to Garward's `85c3af7`, the commit that added Linux bootstrap support, except `GuiScreenDownloadMods.cs.patch`, which Tsu wrote in `95e6513`. `LoadBalancer.cs.patch` isn't client code at all, it's Tsu's own physics-thread fix from PR #158 living in `Vintagestory.Common`. Replied inline on each with the specific commit and what it fixes.

The pattern behind all of them: `VintagestoryLib.dll` is one assembly holding both `Vintagestory.Client` and `Vintagestory.Server`. Getting the server code this repo cares about to decompile and recompile means the client namespace has to compile too, so every one of these patches is fixing a decompiler-to-compiler mismatch (an ambiguous type, a `ref`-cast pattern the decompiler emits that is not legal to write by hand, an `op_Implicit` call, a hash-named regex generator class that moves between ilspycmd runs), not changing what the client does. None of them touch behavior.

Added an eighth commit since you reviewed: finishes the 1.22.5 move Tsu's `d3ed574` started (`StratumInfo.BaseGameVersion` bumped there, `forks.json` never followed), verified the four open-source fork refs need no change since each is already pinned at that repo's current HEAD, re-extracted the whole patch set against the real 1.22.5 archive, and re-ran the full verification (fresh bootstrap, build, headless boot with real chunk generation) on that baseline. First commit's message now explains why it targeted 1.22.4 on purpose and where the 1.22.5 work landed.

tehtelev (approved):
Ok. Looks good

PR #186, opened 2026-07-25
Lock two races cross-stage concurrency exposed: schematic unpack, chunk-db reads

Zaldaryon:
Fixes the `PathwayStarts` crash @tehtelev hit while reviewing #179, plus a **critical data-corruption bug his review of this PR surfaced**, which turns out to be live on `indev` right now.

> [!WARNING]
> **The third fix below is a live bug on current `indev`, not a regression from this PR.** Until this merges, servers on `indev` are permanently corrupting schematic block-code maps. See "The `BlockCodes` destruction" below. Happy to split it into its own PR if that helps it merge faster.

## The `BlockCodes` destruction (new, critical)

@tehtelev reported a warning on this PR he had never seen before (`block ID 5983 not found in the block registry`), a `KeyNotFoundException` on #188 (`The given key '3683' was not present in the dictionary` in `BlockSchematic.PlaceDecors`), and a warning on #191 (`There is already game:block crate/11562, expected microblock`). **All three are one bug.**

Commit `a9719d0`, shipped as #183 and now on `indev`, converted `BlockSchematicStructure.BlockCodesTmpForRemap` from a **per-instance** field to `[ThreadStatic] private static`. `[ThreadStatic]` on a `static` gives one slot per thread shared by **every schematic instance on that thread**; the original was one per schematic. That difference is the whole bug:

```csharp
if (replaceBlocks == null)
{
    stratumBlockCodesTmpForRemap = BlockCodes;   // aliases THIS schematic's permanent map
    return;
}
if (stratumBlockCodesTmpForRemap == null || stratumBlockCodesTmpForRemap == BlockCodes) stratumBlockCodesTmpForRemap = new Dictionary<int, AssetLocation>();
else stratumBlockCodesTmpForRemap.Clear();      // clears whatever it points at
```

The identity guard only compares against **this** instance's `BlockCodes`. It cannot see that the slot currently aliases a *different* schematic's map. So:

1. Schematic **A** places with `replaceBlocks == null`, and the slot now aliases `A.BlockCodes`.
2. Schematic **B** places on the same thread with `replaceBlocks != null`, the slot is non-null and `!= B.BlockCodes`, so it falls through and **clears `A.BlockCodes`**.

`BlockCodes` is the schematic's permanent id to `AssetLocation` map, loaded once and reused for the process lifetime, so this corrupts that schematic permanently. `BlockSchematic.PlaceOneDecor` then does a bare `BlockCodes[id]` indexer on the emptied map, which is exactly the `KeyNotFoundException`.

**No concurrency is needed to trigger it.** `WorldGenStructure.resolvedRockTypeRemaps` starts `null` and is only populated when a structure configures a rock-type remap group, so structures without remaps supply step 1 and structures with them supply step 2. A normal world has both.

**Fix:** `resolveReplaceRemapsForBlockEntities` now returns the map instead of writing a shared field, both call sites hold it in a local, and the `replaceBlocks == null` path returns `BlockCodes` straight through without ever aliasing or mutating it, matching what vanilla does at `BlockSchematic.cs:709`. A thread-static scratch dictionary is kept for the `replaceBlocks != null` case only, preserving #183's allocation-reuse intent; since it can now only ever hold its own scratch instance, clearing it is provably safe.

## The two original races

**`BlockSchematicStructure.Unpack()`** had an unsynchronized check-then-act on `blocksByPos`. `Init()` sets that field before `LoadMetaInformationAndValidate` finishes populating `PathwayStarts`, so a second thread could see it non-null and read fields that do not exist yet. A lock around both overloads closes it, with no unlocked fast path on purpose: `blocksByPos != null` does not mean the unpack finished.

**`SQLiteDbConnectionv2`**'s `ChunkExists` and `GetChunk` never took `transactionLock` while every write path already did. A concurrent write's open transaction can leave a read command's `Transaction` unset, throwing at `ExecuteReader()`. Locking the two reads matches the discipline the write paths already use.

Both are reachable from cross-stage concurrency alone, which has existed since #175's dispatcher model, so neither needs a raised cap. That ordering matters for #191, which **cannot** merge without this one: raising TerrainFeatures' cap without the `Unpack` lock crashes in `WorldGenStructure.TryGenerateUnderground`.

## Verification

Bootstrap clean, two-pass Release build 0 errors, `smoke-test.sh` reaches RunGame, exactly 2 log lines matching `error|exception` and both are the known false positives (a class literally named `ErrorReporter`, and a line reading "no errors").

Honest limit: a smoke test cannot reproduce the corruption, which needs a large concurrent pregen with a mix of remapped and non-remapped structures. The fix is verified by construction (no path can now mutate or clear a `BlockCodes`-derived reference, checked by inspection and grep) rather than by observing the warnings stop. **Confirming that needs a rerun of your own radius-100 test**, which is the main thing I would like from a re-review.

/cc @tehtelev

### DIFF
--- a/patches/VSEssentials/Systems/WorldGen/Standard/Datastructures/BlockSchematicStructure.cs.patch
+++ b/patches/VSEssentials/Systems/WorldGen/Standard/Datastructures/BlockSchematicStructure.cs.patch
@@ -1,5 +1,5 @@
 diff --git a/VSEssentials/Systems/WorldGen/Standard/Datastructures/BlockSchematicStructure.cs b/VSEssentials/Systems/WorldGen/Standard/Datastructures/BlockSchematicStructure.cs
-index 7573f32..27a7943 100644
+index 7573f32..d75ab17 100644
 --- a/VSEssentials/Systems/WorldGen/Standard/Datastructures/BlockSchematicStructure.cs
 +++ b/VSEssentials/Systems/WorldGen/Standard/Datastructures/BlockSchematicStructure.cs
 @@ -12,11 +12,30 @@ using Vintagestory.ServerMods.NoObf;
@@ -123,3 +123,61 @@ index 7573f32..27a7943 100644
              return placed;
          }
  
+@@ -512,29 +536,52 @@ namespace Vintagestory.ServerMods
+             cloned.OriginalPos = OriginalPos;
+ 
+             return cloned;
+         }
+ 
++        // Stratum: guards the lazy unpack below. schematicDatas entries (see WorldGenStructure)
++        // are cached objects reused by every column that draws this schematic, and worldgen has
++        // always let different columns run in different pipeline stages at the same time,
++        // independent of any one stage's own concurrency cap. The original check-then-act let a
++        // second thread observe blocksByPos already non-null, since Init() sets it early, well
++        // before LoadMetaInformationAndValidate finishes populating fields like PathwayStarts,
++        // and skip straight into reading those fields before the thread doing the unpack had set
++        // them. That is the PathwayStarts crash reported while reviewing #179.
++        //
++        // No lock-free fast path on purpose: blocksByPos becoming non-null does not mean
++        // unpacking has finished, so every caller has to go through the lock, not just the one
++        // that finds blocksByPos still null.
++        private readonly object stratumUnpackLock = new object();
++
+         /// <summary>
+         /// Unpack blocks and indices to an array of positioned blocks, and other initialisation steps needed prior to placement and prior to Pathway checks
+         /// <br/>From 1.19.4 we do this Unpack() lazily, only when required, to save RAM [maybe also speeds up game start time, especially if mods add more schematics]
+         /// </summary>
+         /// <param name="api"></param>
+         public void Unpack(ICoreAPI api)
+         {
+-            if (blocksByPos == null)
++            lock (stratumUnpackLock)
+             {
+-                Init(api.World.BlockAccessor);
+-                LoadMetaInformationAndValidate(api.World.BlockAccessor, api.World, FromFile);
++                if (blocksByPos == null)
++                {
++                    Init(api.World.BlockAccessor);
++                    LoadMetaInformationAndValidate(api.World.BlockAccessor, api.World, FromFile);
++                }
+             }
+         }
+ 
+         public void Unpack(ICoreAPI api, int orientation)
+         {
+-            if (orientation > 0 && blocksByPos == null)
++            if (orientation > 0)
+             {
+-                TransformWhilePacked(api.World, EnumOrigin.BottomCenter, orientation * 90, null);
++                lock (stratumUnpackLock)
++                {
++                    if (blocksByPos == null)
++                    {
++                        TransformWhilePacked(api.World, EnumOrigin.BottomCenter, orientation * 90, null);
++                    }
++                }
+             }
+             Unpack(api);
+         }
+     }
+ }
--- a/patches/VintagestoryLib/Vintagestory.Common.Database/SQLiteDbConnectionv2.cs.patch
+++ b/patches/VintagestoryLib/Vintagestory.Common.Database/SQLiteDbConnectionv2.cs.patch
@@ -1,0 +1,68 @@
+diff --git a/VintagestoryLib/Vintagestory.Common.Database/SQLiteDbConnectionv2.cs b/VintagestoryLib/Vintagestory.Common.Database/SQLiteDbConnectionv2.cs
+index 0c59a67..a7ed24d 100644
+--- a/VintagestoryLib/Vintagestory.Common.Database/SQLiteDbConnectionv2.cs
++++ b/VintagestoryLib/Vintagestory.Common.Database/SQLiteDbConnectionv2.cs
+@@ -326,14 +326,23 @@ public class SQLiteDbConnectionv2 : SQLiteDBConnection, IGameDbConnection, IDisp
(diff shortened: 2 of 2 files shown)


tehtelev (changes requested):
I tested it on three world-generation threads.
I ran several tests; there are no errors, but I’m getting notifications like these. I’ve never seen notifications like this before this PR. I don’t think we should ignore this.
`[Warning] Cannot load support beam block ID mapping @ 511573, 113, 512237; block ID 5983 not found in the block registry. Will not display correctly.`
[server-main.log](https://github.com/user-attachments/files/30385478/server-main.log)
[server-worldgen.log](https://github.com/user-attachments/files/30385480/server-worldgen.log)

tehtelev:
A clean test run was performed on build 02a43f69b2ab62c086dd09a211308e9c57b268c1 

[server-main.log](https://github.com/user-attachments/files/30385333/server-main.log)

I compared all the tests and got the following results:
image

Zaldaryon:
Thanks for testing this at 3 threads, that warning was worth chasing. You were right not to ignore it, and it turned out worse than a warning.

**It is one bug, and it explains three of your four reports.** The `block ID 5983 not found` here, the `KeyNotFoundException` you hit on #188, and the `already crate, expected microblock` on #191 all come from the same place.

`a9719d0`, which shipped as #183, converted `BlockSchematicStructure.BlockCodesTmpForRemap` from a per-instance field to `[ThreadStatic] static`. That gives one slot per thread shared by every schematic on that thread, where the original was one per schematic. The identity guard only checks against the current instance's `BlockCodes`, so it cannot tell that the slot is aliasing a *different* schematic's map, and the next placement with rock-type remaps calls `Clear()` on it. `BlockCodes` is that schematic's permanent id-to-code map, so it stays corrupted for the rest of the process. `PlaceOneDecor` then indexes it bare, which is your `KeyNotFoundException` exactly.

Two things worth being blunt about:

1. **This is live on `indev` right now**, not a regression from this PR. It needs no concurrency at all: `resolvedRockTypeRemaps` starts null, so structures without remaps alias the slot and structures with them clear it, and a normal world has both. You saw it here first, but any server on current `indev` is corrupting schematics.
2. **It is mine**, from #183.

Fixed in `aa46ef6`: the remap map is returned and held in a local instead of living in a shared field, and the `replaceBlocks == null` path passes `BlockCodes` straight through without aliasing or mutating it, which is what vanilla does anyway. The thread-static scratch is kept for the remap case only, so it can now only ever hold its own dictionary.

**If you would rather this merged immediately and separately from the `Unpack` lock work, say the word and I will split it into its own PR.** Given it is corrupting data on `indev` I would understand wanting it in first.

What I could not verify: a smoke test cannot reproduce the corruption, so the fix is verified by construction rather than by watching the warnings stop. **A rerun of your radius-100 test is the thing that would actually confirm it**, and it would also tell us whether #191's warning clears, since I expect it has the same cause.

Zaldaryon:
**Raising urgency: `v1.22.5-stratum.1` shipped last night with this bug in it.** Not a criticism of the release, the fix simply was not merged yet, but it changes who is exposed from "people running `indev`" to "everyone who installs the stable build".

Verified against the released `main` (tag points at `59320c1a`):

| Check | Released `main` |
| --- | --- |
| buggy `stratumBlockCodesTmpForRemap` field | present, 8 occurrences |
| fixed `stratumBlockCodesRemapScratch` field | absent |
| the destructive `else ...Clear();` line | present |
| PreDone and TerrainLate concurrency caps | both present |

So the release carries the corruption **and** the two raised caps. The caps are not what makes it trigger, since the destructive sequence is reachable single threaded, but they do mean more schematic placements interleave.

Why this is worse than a log warning: `BlockCodes` is the schematic's permanent id-to-code map, so once it is cleared that schematic is wrong for the rest of the process. The visible symptoms are the ones already reported across these PRs, `block ID not found`, `KeyNotFoundException` in `PlaceDecors`, and block entities resolving to the wrong type. **The output of those paths is written into the world and persists**: wrong blocks placed, structures aborted part-way through decor placement. A player generating fresh chunks on the stable build can end up with permanently malformed ruins, and nothing in the log tells them their world is affected rather than just noisy.

The fix is `aa46ef6` on this branch, and it is small and self-contained: `resolveReplaceRemapsForBlockEntities` returns the map instead of writing a shared field, the `replaceBlocks == null` path passes `BlockCodes` straight through untouched (which is what vanilla does), and the thread-static scratch is kept for the remap case only, where it can now only ever hold its own dictionary.

@tehtelev this branch is still marked changes-requested from your original review, which is what surfaced the bug in the first place. **If it would be faster, I am happy to split just this fix into its own minimal PR against `indev` so it can go out as a patch release without waiting on the `Unpack` lock work in the rest of this branch.** Say which you prefer and I will do it now.

cc @trevorftp since you cut the release and would be the one deciding whether this warrants a `.2`.

Zaldaryon:
The `BlockCodes` corruption fix from this branch went in separately as #195 and is now on `indev`, since it was live in `v1.22.5-stratum.1` and warranted shipping on its own rather than waiting on the rest of this branch.

**Reproduced and verified before merging**, rather than reasoned about as it was here. Radius 100 pregen, 3 worldgen threads, seed `1027995113`: 2 `KeyNotFoundException` hits and 1 worldgen exception before, 0 and 0 after, with the reproduced stack trace matching yours line for line. Pregen wall clock 6m45s against 6m49s, so it costs nothing.

This branch is rebased onto that and is now **narrower**: just the `Unpack` lock (your `PathwayStarts` crash) and the `SQLiteDbConnectionv2` read locks. Two files.

PR #187, opened 2026-07-25
Audit every TerrainFeatures generator, fix the two that crash plus one more landmine

Zaldaryon:
The TerrainFeatures generator audit @tehtelev asked for in the second review pass on #179. #183 already closed the landmines in `GenDeposits`, `GenStructures`, `GenVegetationAndPatches`, and `BlockSchematicStructure`; this finishes the stage by going through every remaining generator registered at `EnumWorldGenPass.TerrainFeatures`.

## What each generator needed

| Generator | Verdict | Fix |
| --- | --- | --- |
| `GenTerraPostProcess` | crashed | per-column scratch to `[ThreadStatic]` |
| `GenPonds` | crashed | per-column scratch to `[ThreadStatic]` |
| `GenDungeons` | two distinct hazards | `[ThreadStatic]` for the Randoms, `ConcurrentDictionary` plus a dedicated lock for genuinely shared region state |
| `GenHotSprings` | already clean | none |

`GenHotSprings` is worth recording explicitly so the next audit does not redo the check: every per-column value there is already a local, and the only shared state it touches is `api.World.Rand`, which `ServerMain` already backs with a `ThreadLocal<Random>`.

`GenDungeons` needed two different fixes because it has two different shapes of state. `rand` and `grassRand` are the familiar per-column-reseeded `LCGRandom` pattern, fixed with `[ThreadStatic]` like everywhere else. `dungeonPlaceTasksByRegion` and `Dungeons` are not per-column scratch at all: they are persistent cross-column, cross-region shared state by design, so `[ThreadStatic]` would have been the wrong tool. The dictionary itself moved to `ConcurrentDictionary`, and a dedicated lock covers the compound read-modify-write sequences that neither a concurrent collection nor per-field atomicity makes safe: region generation's clear-then-repopulate of `Dungeons`, and a column's check-then-compute-then-assign of a shared `DungeonPlaceTask`'s `RockBlockCode` and `SurfacePlaceTasks`. Two columns touching the same dungeon would otherwise silently generate a structure's stairs twice or lose one column's rock-block choice, rather than crash.

## `StratumWorldgenThreadGuard<T>`

Also adds a canary for `[ThreadStatic]`-confined worldgen state, modeled on C2ME-fabric's `CheckedThreadLocalRandom` ([RelativityMC/C2ME-fabric](https://github.com/RelativityMC/C2ME-fabric)).

A `[ThreadStatic]` field cannot be read from the wrong thread by construction, so this should never fire. If it ever does, it means a reference to the guarded value escaped across a thread hop instead of being re-fetched from the field, which is a different and more interesting bug than an unsafe field declaration.

Off by default behind `STRATUM_ENFORCE_SAFE_WORLDGEN_THREAD_ACCESS`. Not wired into `stratum.json` because `StratumConfig`/`StratumRuntime` are internal to `VintagestoryLib` with no `InternalsVisibleTo` toward that assembly. Wired into `GenPonds.rand` as the concrete demonstration rather than retrofitted everywhere in one go.

## Verification

- Release build clean, 0 errors, after fixing one real compile error along the way (`GenDungeons.cs` was missing `using System`, needed for `[ThreadStatic]`).
- `extract-patches.sh` round-trips clean: exactly the four expected new files, nothing unexpected.
- `bash scripts/smoke-test.sh` reaches RunGame with zero fatal errors.
- Pregen comparison, radius 20, `MaxWorldgenThreads=4`: **27s with zero exceptions and zero watchdog fires**, against `indev`'s own clean **24s** baseline. This branch raises no cap, so the useful check was confirming the fixes do not regress anything, and they do not.

One thing worth stating plainly rather than leaving implied: this run also confirmed the "worldgen has become stuck" watchdog seen on `perf/predone-stage-concurrency` at the time (3m52s, one watchdog fire, 178 stuck-column lines, same thread setting) was **not** explained by anything fixed here. Fixing these generators changed nothing about that watchdog. It was root-caused separately, to a same-request double-claim in the dispatcher, and is fixed on the cap-raise branches.

## Relationship to the other branches

This raises no concurrency cap. It is a prerequisite for `perf/terrainfeatures-stage-concurrency`, which does, and which also needs #186 for reasons that branch's description covers.

/cc @tehtelev

---

## Added after @tehtelev's review: `GenPonds` null-chunk guard

His testing of #186+#187 caught one error, a `NullReferenceException` in `GenPonds.TryPlacePondAt` at the TerrainFeatures pass with 3 worldgen threads. Fixed here.

Pond search walks outward from a candidate position and pulls in whatever chunk covers each visited coordinate:

```csharp
chunk = (IServerChunk)blockAccessor.GetChunk(curChunkX, pondYPos / chunksize, curChunkZ);
if (chunk == null) chunk = api.WorldManager.GetChunk(curChunkX, pondYPos / chunksize, curChunkZ);
chunk.Unpack();                       // NRE when the fallback also returns null
```

The first lookup is null checked, the fallback is not. Four lines below, the sibling lookup for `chunkOneBlockBelow` **already** guards with `if (chunkOneBlockBelow == null) return;`, so the omission is an inconsistency rather than a deliberate invariant, and matching it needs no new policy: bailing out abandons one pond placement, which is exactly what the existing guard does.

**Vanilla-inherited, not introduced by this audit.** `.baseline/.../GenPonds.cs` is byte-identical at that spot and this branch's `GenPonds.cs.patch` does not touch the line. What changes is reachability: pond search crosses into neighbouring columns, and once several columns run TerrainFeatures at once a neighbour can legitimately be absent or mid-unload, which is effectively unreachable while the stage stays at one column at a time.

Verified only what a smoke test can show: build clean, server reaches RunGame, exactly the 2 known false-positive log lines. A smoke test cannot reproduce the original crash, so the guard is reasoned from your stack trace rather than observed to stop it. A rerun of your radius-100 test is what would confirm it.

### DIFF
--- a/patches/VSEssentials/Systems/WorldGen/Standard/ChunkGen/1.GenTerra/GenTerraPostProcess.cs.patch
+++ b/patches/VSEssentials/Systems/WorldGen/Standard/ChunkGen/1.GenTerra/GenTerraPostProcess.cs.patch
@@ -1,0 +1,90 @@
+diff --git a/VSEssentials/Systems/WorldGen/Standard/ChunkGen/1.GenTerra/GenTerraPostProcess.cs b/VSEssentials/Systems/WorldGen/Standard/ChunkGen/1.GenTerra/GenTerraPostProcess.cs
+index 79c481d..691b108 100644
+--- a/VSEssentials/Systems/WorldGen/Standard/ChunkGen/1.GenTerra/GenTerraPostProcess.cs
++++ b/VSEssentials/Systems/WorldGen/Standard/ChunkGen/1.GenTerra/GenTerraPostProcess.cs
+@@ -13,12 +13,24 @@ namespace Vintagestory.ServerMods
+     public class GenTerraPostProcess : ModStdWorldGen
+     {
+         ICoreServerAPI api;
+         IWorldGenBlockAccessor blockAccessor;
+ 
+-        HashSet<int> chunkVisitedNodes = new HashSet<int>();
+-        List<int> solidNodes = new List<int>(40);
++        // Stratum: every field below used to be a plain instance field, written
++        // and read within the processing of a single column (OnChunkColumnGen ->
++        // deletePotentialFloatingBlocks -> AddToChunkVisitedNodesIfSameChunk),
++        // unsafe once TerrainFeatures' same-stage cap moves off 1. Matches the
++        // InvalidOperationException: Collection was modified crash reported
++        // against this class: two columns concurrently mutating the same shared
++        // HashSet/List. [ThreadStatic] since GenTerraPostProcess is a singleton
++        // ModSystem with no second live instance anywhere in the codebase
++        // (confirmed by search).
++        [ThreadStatic] private static HashSet<int> stratumChunkVisitedNodes;
++        [ThreadStatic] private static List<int> stratumSolidNodes;
++        [ThreadStatic] private static QueueOfInt stratumBfsQueue;
++        [ThreadStatic] private static int[] stratumCurrentVisited;
++        [ThreadStatic] private static int stratumIteration;
+ 
+         public override bool ShouldLoad(EnumAppSide side)
+         {
+             return side == EnumAppSide.Server;
+         }
+@@ -56,10 +68,11 @@ namespace Vintagestory.ServerMods
+             const int chunksize = GlobalConstants.ChunkSize;
+             const int chunksizeSquared = chunksize * chunksize;
+             int chunkY = seaLevel / chunksize;
+             int yMax = chunks[0].MapChunk.YMax;
+             int cyMax = Math.Min(yMax / chunksize + 1, api.World.BlockAccessor.MapSizeY / chunksize);
++            HashSet<int> chunkVisitedNodes = stratumChunkVisitedNodes ??= new HashSet<int>();
+             chunkVisitedNodes.Clear();
+ 
+             for (int cy = chunkY; cy < cyMax; cy++)
+             {
+                 IChunkBlocks chunkdata = chunks[cy].Data;
+@@ -106,28 +119,29 @@ namespace Vintagestory.ServerMods
+             }
+         }
+ 
+ 
+ 
+-        QueueOfInt bfsQueue = new QueueOfInt();
+         const int ARRAYSIZE = 41;  // Note if this constant is increased beyond 64, the bitshifts for compressedPos in the bfsQueue.Enqueue() and .Dequeue() calls may need updating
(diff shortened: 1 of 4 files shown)


tehtelev (changes requested):
I tested 186 and 187 together with different numbers of threads. I caught only one error.

```
26.7.2026 10:58:08 [Worldgen] An error was thrown in pass TerrainFeatures when generating chunk column X=16066,Z=16072 in world 'Awesome Kingdom Valley' with seed 1027995113
Exception System.NullReferenceException: Object reference not set to an instance of an object.
   at Vintagestory.ServerMods.GenPonds.TryPlacePondAt(Int32 dx, Int32 pondYPos, Int32 dz, Int32 chunkX, Int32 chunkZ, Int32 depth) in C:\Users\tehtelev\Source\Repos\Stratum\VSEssentials\Systems\WorldGen\Standard\ChunkGen\7.GenPonds\GenPonds.cs:line 298
   at Vintagestory.ServerMods.GenPonds.OnChunkColumnGen(IChunkColumnGenerateRequest request) in C:\Users\tehtelev\Source\Repos\Stratum\VSEssentials\Systems\WorldGen\Standard\ChunkGen\7.GenPonds\GenPonds.cs:line 152
   at Vintagestory.Server.ServerSystemSupplyChunks.runGenerators(ChunkColumnLoadRequest chunkRequest, Int32 forPass) in C:\Users\tehtelev\Source\Repos\Stratum\baseline\VintagestoryLib\Vintagestory.Server\ServerSystemSupplyChunks.cs:line 2403
```

[server-main.log](https://github.com/user-attachments/files/30384486/server-main.log)

Zaldaryon:
Fixed in `cc557ca`.

Your stack trace pointed straight at it: `TryPlacePondAt` null checks the first chunk lookup, falls back to `api.WorldManager.GetChunk`, and then calls `chunk.Unpack()` without checking the fallback. Four lines below, the sibling lookup for `chunkOneBlockBelow` already returns early on null, so this was an inconsistency rather than an invariant, and I matched it.

Vanilla-inherited rather than something the audit introduced: `.baseline` is byte-identical at that spot and this branch's patch does not touch the line. Raising the TerrainFeatures cap is what makes it reachable, since pond search crosses into neighbouring columns and a neighbour can legitimately be absent or mid-unload once several columns run the stage at once.

What I verified: build clean, server reaches RunGame, exactly the 2 known false-positive log lines. What I did **not** verify: a smoke test cannot reproduce the original crash, so this is reasoned from your trace rather than observed to be fixed. Your radius-100 run at 3 threads is what would confirm it.

Also worth knowing before you retest: the warning you flagged on #186 turned out to be a data-corruption bug that is live on `indev` and explains three of your four reports across these PRs. Details in the #186 comment. Since you test these stacked on #186+#187, it is worth picking up that fix before the next round.

PR #188, opened 2026-07-25
Let Terrain run several columns at once, with a two-axis mod-safety fallback

Zaldaryon:
> [!CAUTION]
> **Correction, 2026-07-27: this branch did not contain the cap raise its title describes, and every measurement below was taken without it.**
>
> `stratumStageMaxConcurrent[StratumWorldGenStages.Terrain]` was assigned nowhere in this branch, and `git log -S` over `patches/` finds no commit in its whole history that added it. Commit `7240095`, whose message reads "raise Terrain's stage-concurrency cap", changed exactly one file: `GenTerra.cs.patch`. Only the GenTerra half ever landed. That message is also wrong about PreDone and TerrainLate sitting at cap 1 here — both are raised, inherited from `e943289`.
>
> **What this means for the numbers below:** the radius-40 comparisons, including the ~25% figure, measured the GenTerra changes alone against `indev`. They are not evidence about running Terrain concurrently, because Terrain was not running concurrently. Please disregard them as a cap-raise result. The stage-share table and the mod-safeguard design discussion are unaffected.
>
> **What changed in the branch now:**
> - the missing cap line was added, so the branch finally does what it says;
> - `EnforceStageConcurrencySafety` was extended to Terrain. It covered only PreDone and TerrainLate, and Terrain is the pass that most needs it — Rivers and VSRiverGen both keep shared mutable fields on `NewGenTerra`. Without this it would have been the one cap raise shipping with no mod safeguard;
> - the branch was rebased onto #199, which fixes a claim leak in `TryClaimStage` that any raised cap can hit.
>
> **Consequence for the sibling branches:** the safeguard block is no longer byte-identical across all four cap-raise branches, since this one now calls it three times rather than twice. Whichever merges first, the others need that hunk resolved by hand instead of trivially.
>
> **Verification of the corrected branch**, radius 100, seed 1027995113, queue 2000, 3 worldgen threads: 31417/31417 columns, no queue overflow, no stuck columns, no watchdog throws, and the safeguard correctly silent with no mods present, confirming the cap was genuinely active rather than quietly dropped to serial.
>
> **No performance claim is being made to replace the retracted one.** Single runs on this machine did not order the configurations consistently — Terrain at cap 4 measured slower than cap 1 in one pairing and faster in another — so run-to-run spread swallows the effect at n=1. Interleaved same-seed sampling is owed before any figure goes back in this body.
>
> **One behaviour change to disclose:** the cap raise produces `Tried to get block outside generating chunks!` notifications where clean `indev` produces none (286 in the radius-100 run above). Every one comes from a stage still at cap 1, through vanilla code that reads across chunk boundaries — `GenVegetationAndPatches` via `BlockPatch`, `GenTerraPostProcess`, and `GenStructures`. All are reads; there are **no** out-of-range writes in any run, so this is not the `ChunkDataLayer` write-race class holding up #191. `GetBlockId` returns air there, so the visible effect is a skipped or displaced loose rock. Most likely a differently paced frontier letting a neighbour column be evicted before it is read. Not root-caused further.

---

Terrain is one of the two worldgen stages the neighbour-parity gate does **not** hold back, so it is one of the two where raising a concurrency cap can actually bind. This lets it run several columns at once, after making `GenTerra` safe for that.

> [!IMPORTANT]
> **Rewritten after @tehtelev's review.** The `GenTerra` thread-budget divisor this PR originally added turned out to be the entire regression he measured, not a mitigation for it. It has been removed, and with it the new public API member it needed. Details below.

## `GenTerra`'s concurrency-safety fixes

`GenTerra` kept per-column scratch in shared instance fields, the same shape found in every other generator this initiative touched. Fixed with a per-column buffer pool, a `ConcurrentDictionary` for the landform cache, and a double-checked lazy-init lock. That part is unchanged.

## The divisor, and why it is gone

`GenTerra` is the only generator with an **internal `Parallel.For`**. The original theory here was that raising Terrain's cap lets N columns each spawn that fan-out at once, oversubscribing the CPU N times over, so the reserved thread budget should be divided by N. That is what this PR shipped, via a new `IServerAPI.GetWorldgenPassConcurrency`.

It was wrong, and measurably so. `reservedThreadsBudget` is `ProcessorCount` minus 4 or 6, clamped, so a 6 core machine gets **2**. Dividing 2 by a concurrency of 3 floors to **1**, which does not trim the fan-out, it **deletes** it, and every column then generates its whole 32x32 area single threaded. Losing all intra-column parallelism costs far more than the oversubscription it avoids, because those columns interleave on the same cores either way.

Four configurations, radius 40, 3 worldgen threads, same seed, 3+ samples each:

| Configuration | Samples (s) | Median |
| --- | --- | --- |
| clean `indev`, cap 1 | 78, 90, 99 | 90.0s |
| this PR as reviewed, divide by cap | 78, 90, 91 | 90.0s |
| divide by *live* concurrency instead | 87, 96 | 91.5s |
| **no divisor at all** | 60, 60, 66, 69, 69, 72 | **67.5s** |

Only the last separates, and it separates cleanly: every no-divisor run is at or below 72s, every other run at or above 78s. That is about **25% faster than `indev`**.

The middle row is worth recording. Dividing by live concurrency rather than the configured cap was the first fix attempted, on the reasoning that a cap is a ceiling rather than a measurement. It changed nothing, because when Terrain is the bottleneck the live count sits at the cap almost always. That null result is what pointed at the divisor itself.

**Consequence:** `IServerAPI.GetWorldgenPassConcurrency` now has no caller and is removed, along with its `ServerAPI` implementation and the internal `GetStageConcurrency` accessor. This PR now adds **no new public API surface at all**.

## On the reported +10.7%

At radius 40 on my machine, this PR as reviewed and clean `indev` could not be told apart: both 90.0s median, spreads of 12-21s. The +10.7% came from one run per build, and a spread that wide swallows it, so I do not think it was an established regression on its own.

The Terrain ms/col jump was real though, and reproduces here: 10.9-14.7 on `indev` versus 20.1-26.1 with the divisor. That number measures **per-column latency**, not throughput, which is exactly why the divisor looked harmless on the summed per-stage metric while costing 25% of wall clock. Finding it needed the wall-clock comparison, so the report was what surfaced it.

## The mod-safety fallback

| Axis | Catches | How |
| --- | --- | --- |

```
```

| --- | --- | --- | --- |

| --- | --- | --- | --- |

### Honest framing

## Verification

Three mod-safety cases end to end at `MaxWorldgenThreads=4`, with a real minimal test mod (own assembly, references `0Harmony`, dropped into `Mods/` for these tests only):

| Case | Expected | Result |
| --- | --- | --- |
| Mod registers its own Terrain handler | fall back, name the assembly | warned naming `SafeguardTestMod`, cap 1 |
| Mod Harmony-prefixes `GenTerra.OnChunkColumnGen` | fall back, name mod + kind + method | warned naming `safeguardtestmod`, prefix, `GenTerra.OnChunkColumnGen` |

Plus 3 repeated boot trials at `MaxWorldgenThreads=4` (0 stuck-column lines, 0 watchdog throws), a clean build and smoke test after the divisor removal, and the 6 pregen runs above. Every run carried exactly the 2 known false-positive log lines (a class literally named `ErrorReporter`, and a line reading "no errors") and nothing else.

Fixture and harness: `research/atlas-tests/worldgen-safeguard-test-mod/` in the contributions repo.

## Merge ordering

Independent of the other cap-raise branches; merges cleanly into `indev`. The shared `dispatchClaimed` and safeguard blocks are identical across all four, so the rest resolve trivially against whichever lands first.

/cc @tehtelev


tehtelev (changes requested):
I ran tests 186–188 together using different numbers of threads. No critical errors were found.
Only one such error occurred, which may be related to 186.
I also noticed that the generation process appears to have slowed down. 
You can check the column generation time in the logs yourself.



```
26.7.2026 11:14:16 [Worldgen] An error was thrown in pass TerrainFeatures when generating chunk column X=16036,Z=15952 in world 'Awesome Kingdom Valley' with seed 1027995113
Exception System.Collections.Generic.KeyNotFoundException: The given key '3683' was not present in the dictionary.
   at Vintagestory.API.Common.BlockSchematic.PlaceDecors(IBlockAccessor blockAccessor, BlockPos startPos) in C:\Users\tehtelev\Source\Repos\Stratum\VintagestoryApi\Common\Collectible\Block\BlockSchematic.cs:line 725
   at Vintagestory.ServerMods.BlockSchematicStructure.PlaceRespectingBlockLayers(IBlockAccessor blockAccessor, IWorldAccessor worldForCollectibleResolve, BlockPos startPos, Int32 climateUpLeft, Int32 climateUpRight, Int32 climateBotLeft, Int32 climateBotRight, Dictionary`2 replaceBlocks, Int32[] replaceWithBlockLayersBlockids, Boolean replaceMetaBlocks, Boolean replaceBlockEntities, Boolean suppressSoilIfAirBelow, Boolean displaceWater) in C:\Users\tehtelev\Source\Repos\Stratum\VSEssentials\Systems\WorldGen\Standard\Datastructures\BlockSchematicStructure.cs:line 318
   at Vintagestory.ServerMods.WorldGenStructure.TryGenerateAtSurface(IBlockAccessor blockAccessor, IWorldAccessor worldForCollectibleResolve, BlockPos startPos, String locationCode) in C:\Users\tehtelev\Source\Repos\Stratum\VSEssentials\Systems\WorldGen\Standard\ChunkGen\6.GenStructures\WorldGenStructure.cs:line 786
   at Vintagestory.ServerMods.GenStructures.DoGenStructures(IMapRegion region, Int32 chunkX, Int32 chunkZ, Boolean postPass, String locationCode, ITreeAttribute chunkGenParams) in C:\Users\tehtelev\Source\Repos\Stratum\VSEssentials\Systems\WorldGen\Standard\ChunkGen\6.GenStructures\GenStructures.cs:line 302
   at Vintagestory.ServerMods.GenStructures.OnChunkColumnGenPostPass(IChunkColumnGenerateRequest request) in C:\Users\tehtelev\Source\Repos\Stratum\VSEssentials\Systems\WorldGen\Standard\ChunkGen\6.GenStructures\GenStructures.cs:line 235
   at Vintagestory.Server.ServerSystemSupplyChunks.runGenerators(ChunkColumnLoadRequest chunkRequest, Int32 forPass) in C:\Users\tehtelev\Source\Repos\Stratum\baseline\VintagestoryLib\Vintagestory.Server\ServerSystemSupplyChunks.cs:line 2653
```

[server-main.log](https://github.com/user-attachments/files/30384691/server-main.log)

Zaldaryon:
You were right that something was wrong here, and it was worse than a tuning problem: **the divisor this PR added was the regression, not a mitigation for it.** Removed in `4009b01`, and the PR body is rewritten.

The theory was that raising Terrain's cap lets N columns each spawn `GenTerra`'s internal `Parallel.For`, so the thread budget should be divided by N. The flaw is that `reservedThreadsBudget` is `ProcessorCount` minus 4 or 6, clamped, so a 6 core box gets **2**. Dividing 2 by 3 floors to **1**, which does not trim the fan-out, it deletes it, and every column generates its whole 32x32 area single threaded.

Radius 40, 3 threads, same seed, stray build daemons killed before each run:

| Configuration | Samples (s) | Median |
| --- | --- | --- |
| clean `indev`, cap 1 | 78, 90, 99 | 90.0s |
| this PR as you reviewed it | 78, 90, 91 | 90.0s |
| divide by *live* concurrency instead | 87, 96 | 91.5s |
| **no divisor at all** | 60, 60, 66, 69, 69, 72 | **67.5s** |

Only the last separates, and cleanly: every no-divisor run at or below 72s, every other run at or above 78s. About **25% faster than `indev`**.

The third row is why I trust the fourth. Dividing by live concurrency instead of the configured cap was my first attempt, and it changed nothing, because when Terrain is the bottleneck the live count sits at the cap anyway. That null is what pointed at the divisor itself rather than at how it was computed.

Since the divisor is gone, `IServerAPI.GetWorldgenPassConcurrency` has no caller and is removed with its implementation and internal accessor. **This PR now adds no new public API surface at all**, which I think is a straight improvement to its reviewability.

**On the +10.7%, and I want to be careful here because your report is what found this.** At radius 40 I could not separate this PR as it stood from clean `indev`: both 90.0s median, spreads of 12-21s. Your figure came from one run per build, and a spread that wide swallows it. So I do not think that specific number was an established regression on its own.

The Terrain ms/col jump you saw is real though, and reproduces here: 10.9-14.7 on `indev` against 20.1-26.1 with the divisor. It measures per-column latency rather than throughput, which is exactly why the divisor looked harmless on the summed per-stage metric while costing real wall clock. That is a useful thing to know about that metric generally: summing per-stage ms/col over-counts when stages run concurrently, so it is good for localising a regression to a stage and misleading as a total. Localising it to Terrain is what your table did, and that is what made this findable.

If you have the time, repeated runs on this one specifically would be worth more than a single run on all six.

Zaldaryon:
Rebased onto current `indev` now that #190 and #189 have merged, so this shows no conflicts for the re-review.

Both of those carried the shared dispatcher infrastructure this branch also carried independently (the per-stage concurrency arrays, `dispatchClaimed`, and the two-axis mod-safety fallback), exactly as the PR bodies said would happen. All of that now comes from `indev`, so this branch adds only what is specific to its own stage.

Resolution was mechanical and done at the source level then re-extracted, never by hand-editing patch text: conflicting hunks took `indev`'s side, then this stage's own two lines were re-added on top, its cap raise and its `EnforceStageConcurrencySafety` call, since taking `indev`'s side alone would have silently dropped them. Verified after the rebase: build 0 errors, `smoke-test.sh` reaches RunGame, exactly the 2 known false-positive log lines.

Zaldaryon:
**Holding this one: it stalls at radius 100.** Not merging it with the others.

I merged the non-cap-raising PRs into `indev` (#195, #196, #186, #187, #194) and validated the result at radius 100 with your seed: clean, zero worldgen exceptions, zero beam-mapping warnings, zero overlapping-ruin warnings. Then I validated this branch on top of that, same seed and radius, and it does not finish:

```
pregen progress: running 35.32% elapsed=25m 0s eta=45m 48s
pregen progress: running 35.32% elapsed=26m 0s eta=47m 37s
pregen progress: running 35.32% elapsed=27m 0s eta=49m 27s
```

Frozen at 35.32% while the server sits at about 52% CPU, so it is spinning rather than deadlocked idle. No watchdog fire, no stuck-column lines, no worldgen exceptions. It simply stops making progress.

**Scale-dependent, which is why I did not catch it earlier.** At radius 40 this branch was the fastest configuration I measured, 60-72s against 78-99s for `indev`. At radius 100 it never completes. Every number I posted for this PR came from radius 40, so the 25% figure applies to that scale only and should not be read as a general claim.

Not root-caused yet. I did not want to sit on the finding while investigating, since the merge queue was moving.

Same call for #191, for a different reason; details there.

PR #189, opened 2026-07-25
Let TerrainLate run several columns at once, with a two-axis mod-safety fallback

Zaldaryon:
TerrainLate is the second-heaviest stage, 21.9% of pregen time. This lets it run several columns at once, after fixing its four generators for it.

## The four generators

TerrainLate is Stratum's own second half of the split Terrain pass, so everything here registers through `StratumTerrainSplit` rather than the vanilla pass API.

| Generator | Fix |
| --- | --- |
| `GenRockStrata` | per-column scratch to thread-local; region province cache `Dictionary` → `ConcurrentDictionary` |
| `GenCaves` | per-column-reseeded `LCGRandom` to thread-local |
| `GenBlockLayers` | per-column scratch to thread-local |
| `GenDevastationLayer` | none needed, already call-local or read from the per-request `chunks` array |

Each had its own version of the same hazard: per-column scratch, or a per-column-reseeded `Random`, living in a shared instance field, written in one method and read a few calls later in another, all within one column's processing.

`GenRockStrata`'s province cache needed a different treatment from the rest. It is not per-column scratch but a genuine region-level cache, and once two columns from different, not-yet-cached regions can run together they can legitimately race on it, so it became a `ConcurrentDictionary` rather than moving to thread-local storage.

`GenCaves` and `GenRockStrata` use `ThreadLocal` instance fields rather than `[ThreadStatic]` statics, because they extend `GenPartial`, which has more than one live instance: `ProPickWorkSpace` constructs a second one for prospecting-pick probing. A static field would leak across those instances.

## The mod-safety fallback

| Axis | Catches | How |
| --- | --- | --- |

```
```

| --- | --- | --- | --- |

| --- | --- | --- | --- |

### Honest framing

## One detail specific to this stage

Three of these generators do not declare their own handler; they inherit `GenChunkColumn` from `GenPartial`. A delegate over an unoverridden virtual reports the declaring type, so `handler.Method` resolves to `GenPartial.GenChunkColumn`, which is also the only method Harmony will let a mod patch there.

That was confirmed rather than assumed. Trying to patch `GenCaves.GenChunkColumn` is rejected by Harmony itself ("You can only patch implemented methods/constructors"), and patching the declared `GenPartial.GenChunkColumn` is both what a real mod has to do and what the safeguard catches.

## Verification

Three cases, end to end, at `MaxWorldgenThreads=4`, with a real minimal test mod built for this (own assembly, references `0Harmony`, dropped into `Mods/` for these tests only):

| Case | Expected | Result |
| --- | --- | --- |
| Mod registers own handler via `StratumTerrainSplit` | fall back, name the assembly | warned naming `SafeguardTestMod`, cap 1 |
| Mod Harmony-prefixes `GenPartial.GenChunkColumn` | fall back, name mod + kind + method | warned naming `safeguardtestmod`, prefix, `GenPartial.GenChunkColumn` |

All three reached RunGame with zero real crashes and the same 2 known false-positive log lines each (a class literally named `ErrorReporter`, and a line reading "no errors").

Also 3 repeated boot trials at `MaxWorldgenThreads=4`: 0 stuck-column lines, 0 watchdog throws. Rebased onto current `indev` and re-verified after rebase: bootstrap clean, build 0 errors, smoke test reaches RunGame.

## Merge ordering

Independent of the other three cap-raise branches; merges cleanly into `indev` today. The shared `dispatchClaimed` and safeguard blocks are identical across all four, so the rest resolve trivially against whichever lands first.

/cc @tehtelev

### DIFF
--- a/patches/VSEssentials/Systems/WorldGen/Standard/ChunkGen/2.GenRockStrata/GenRockStrata.cs.patch
+++ b/patches/VSEssentials/Systems/WorldGen/Standard/ChunkGen/2.GenRockStrata/GenRockStrata.cs.patch
@@ -1,8 +1,34 @@
 diff --git a/VSEssentials/Systems/WorldGen/Standard/ChunkGen/2.GenRockStrata/GenRockStrata.cs b/VSEssentials/Systems/WorldGen/Standard/ChunkGen/2.GenRockStrata/GenRockStrata.cs
-index cb9eefa..c3f1057 100644
+index cb9eefa..dcb6063 100644
 --- a/VSEssentials/Systems/WorldGen/Standard/ChunkGen/2.GenRockStrata/GenRockStrata.cs
 +++ b/VSEssentials/Systems/WorldGen/Standard/ChunkGen/2.GenRockStrata/GenRockStrata.cs
-@@ -46,11 +46,11 @@ namespace Vintagestory.ServerMods
+@@ -1,7 +1,9 @@
+ using System;
++using System.Collections.Concurrent;
+ using System.Collections.Generic;
++using System.Threading;
+ using Vintagestory.API.Common;
+ using Vintagestory.API.Datastructures;
+ using Vintagestory.API.MathTools;
+ using Vintagestory.API.Server;
+ using Vintagestory.ServerMods.NoObf;
+@@ -24,11 +26,14 @@ namespace Vintagestory.ServerMods
+         internal SimplexNoise distort2dx;
+         internal SimplexNoise distort2dz;
+         internal MapLayerCustomPerlin[] strataNoises;
+ 
+         int regionMapSize;
+-        Dictionary<int, LerpedWeightedIndex2DMap> ProvinceMapByRegion = new Dictionary<int, LerpedWeightedIndex2DMap>(10);
++        // Stratum: ConcurrentDictionary since TerrainLate can now run concurrently for
++        // distant columns (see StratumWorldGenStages), and two columns from different,
++        // not-yet-cached regions could otherwise race on the same bucket at once.
++        ConcurrentDictionary<int, LerpedWeightedIndex2DMap> ProvinceMapByRegion = new ConcurrentDictionary<int, LerpedWeightedIndex2DMap>();
+ 
+         public override bool ShouldLoad(EnumAppSide side)
+         {
+             return side == EnumAppSide.Server;
+         }
+@@ -46,11 +51,11 @@ namespace Vintagestory.ServerMods
          public override void StartServerSide(ICoreServerAPI api)
          {
              this.api = api;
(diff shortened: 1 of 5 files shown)


tehtelev (approved):
Tests were run with 186, 187, and 189 with varying numbers of threads.
No issues were found with this PR.

[server-main.log](https://github.com/user-attachments/files/30384896/server-main.log)

Zaldaryon:
No code change here, thanks for the approval. Flagging two things from the other reviews that affect this one, since you tested it stacked on #186+#187.

**Pick up #186 before merging this.** The warning you flagged there turned out to be a data-corruption bug that is live on `indev` today: a `[ThreadStatic] static` remap map shared across schematic instances ends up clearing a different schematic's permanent `BlockCodes`. It needs no concurrency to trigger and it explains three of your four reports across these PRs. Fixed in `aa46ef6`.

**The performance section in this body has been restated**, because two claims in it were wrong:

- The stage shares came from a radius-20 profile. At your radius-100 scale TerrainLate is 29.9% and Vegetation 27.7% while Terrain is 17.9%, not the 42.7% I originally quoted.
- Your Discord point about neighbour parity checks out, and I instrumented it. `ensurePrettyNeighbourhood` returns true immediately at or below TerrainLate and otherwise requires all 8 neighbours at the same pass. Measured across every run: Terrain and TerrainLate block exactly 0 times, the gated stages block heavily. So the honest framing is that the value is concentrated in the two ungated stages, and caps on gated stages are latent capability rather than throughput.

I also measured your queue hypothesis and it did not hold: radius 60, 4 interleaved samples per config, queue=2000 median 170.5s against queue=8000 median 170.0s, with 34-39s spreads. The exhaustion warnings do disappear, they just do not buy throughput. Worth knowing that every earlier radius-20 measurement in this initiative ran 1,681 columns against a 2,000 slot queue, so none of them could ever have stressed it, which is why I never saw what you saw.

Zaldaryon:
Heads up before this merges: **I force-pushed a rebase onto the new `indev`**, so the commits you approved have different hashes now. Flagging it rather than sliding it past you.

#190 merged first, and it carried the shared dispatcher work (`dispatchClaimed`, the per-stage concurrency arrays, the mod-safety fallback) that this branch also carried independently, exactly as the PR bodies said would happen. So the rebase dropped three of this branch's four commits as already-applied and kept one.

Resolution was mechanical, and done at the source level then re-extracted rather than by hand-editing patch text: every conflicting hunk took `indev`'s side, then TerrainLate's own two lines were re-added on top, its cap raise and its `EnforceStageConcurrencySafety` call, since taking `indev`'s side alone would have silently dropped them.

Verified after the rebase: build 0 errors, `smoke-test.sh` reaches RunGame, exactly the 2 known false-positive log lines. Content check against `indev` shows what this branch should add and nothing else: the four TerrainLate generator patches (`GenRockStrata`, `GenCaves`, `GenBlockLayers`, `GenPartial`) plus the two dispatcher lines. Both cap raises now coexist, PreDone's from `indev` and TerrainLate's from here.

Your approval is still recorded and the content is equivalent, so I am merging on it. **If you would rather re-check the rebased version first, say so and I will revert the merge.**

Zaldaryon:
**I need to flag a regression I introduced by merging this, and it disagrees with your own measurement, which is the interesting part.**

After merging #190 and then this, I checked that `indev` still built and booted, and then measured it. Radius 40, 3 worldgen threads, same seed, stray build daemons killed before each run, distinct data dir per run, on a **6 core** machine:

| `indev` state | Samples (s) | Median |
| --- | --- | --- |
| pre-merge (`7e8fefe`) | 78, 82, 90, 91, 99 | 90s |
| + #190 only (`33e947c`) | 82, 99 | ~90s |
| **+ #190 + #189 (`e943289`)** | 117, 130, 142 | **130s** |

No overlap: the slowest pre-merge run is 99s, the fastest post-merge run is 117s. So roughly **+44%**, and it isolates cleanly to this PR rather than #190, which is what I would expect given PreDone is 0.2% of stage time.

**Your radius-100 numbers say the opposite**, and I do not think either of us measured wrong. Your run had #186+187+189 at 234s against a clean 234s, dead level. The likely difference is core count. `TerrainLate`'s cap is set to `additionalWorldGenThreadsCount`, which comes from `MaxWorldgenThreads` and **does not consider how many cores the machine actually has**. On 6 cores with 3 worldgen threads, plus the ticking thread, chunkdbthread, pathfinding and the rest, putting 3 columns into TerrainLate at once oversubscribes badly. On a larger box it would not.

Worth noting this is the same shape as what I found on #188: raising an ungated stage's cap is not uniformly good, and it needs measuring per stage and apparently per machine. The difference there was that `GenTerra` has an internal `Parallel.For`, so one column already spreads across threads and the trade is visible; TerrainLate's four generators have no internal fan-out, so raising its cap adds concurrent columns with nothing to trade against, and it lands as pure contention.

I verified the merge itself is correct, not botched: all four TerrainLate generator patches are present on `indev` and their thread-safety changes are in the built source, and a radius-40 pregen shows 0 stuck columns and 0 watchdog fires. It is correct, just slower here.

**What I would suggest, but it is your call as maintainer:**

1. Revert this merge on `indev` while we work out the cap, since a 44% hit on a 6 core box is a realistic small-server configuration, or
2. Keep it and make the cap core-aware, so it does not exceed what the machine can actually absorb. That is a small change and it would also protect #188 and #191 from the same trap.

I would lean toward 2 with a revert in the meantime, but I am not going to revert a shared branch unilaterally. Tell me which you want and I will do it. If it helps, your core count and `MaxWorldgenThreads` on the test box would let me reproduce your side rather than guessing at the difference.

Zaldaryon:
**Retracting my previous comment: there is no regression. That measurement was contaminated and I should not have posted it.**

I re-measured the exact same commit (`e943289`) from a clean state, and also current `indev`:

| `indev` state | Samples (s) | Median |
| --- | --- | --- |
| pre-merge (`7e8fefe`) | 78, 82, 90, 91, 99 | 90s |
| `e943289`, re-measured clean | 81, 97 | 89s |
| current `indev` (`9574d8f`) | 87, 91, 96 | 91s |
| my earlier claim for `e943289` | 117, 130, 142 | ~~130s~~ |

Every honest reading sits in the high 80s to low 90s. The 117-142 readings do not reproduce.

What went wrong: I took those samples immediately after a long sequence of bootstraps, builds and back-to-back pregens, without a contemporaneous clean-state control. This project has already been burned once by leftover build processes faking a 4.4x regression, and I knew that, and I still published a conclusion from a run series that had the same defect. I took multiple samples, which is why I trusted it, but multiple contaminated samples are still contaminated.

So concretely:

- **Do not revert this merge.** My previous comment asked you to consider that. Disregard it.
- **Your radius-100 result was right**, and my disagreement with it was the artefact, not your number. The two of us were not measuring different machines behaving differently; I was measuring a dirty machine.
- The core-awareness observation stands on its own merits as a design point, since the caps genuinely never consult `Environment.ProcessorCount`, but it is **not** motivated by an observed regression and should not be prioritised as if it were.

Apologies for the noise on a merged PR.

PR #190, opened 2026-07-25
Let PreDone run several columns at once, with a two-axis mod-safety fallback

Zaldaryon:
PreDone was the last worldgen stage still running one column at a time. This raises its cap, after making its single generator safe for it.

## `GenCreatures`

PreDone hosts exactly one generator, and nothing had audited whether it tolerates two columns at once. It does not. Climate, forest, and shrub readings live in plain instance fields that `OnChunkColumnGen` writes and `TrySpawnGroupAt` reads a few calls later, so a second thread starting a different column mid-flight overwrites the first thread's readings before it is done with them. The shared spawn-position list and the single world-lifetime `Random` have the same problem.

Every piece of per-column scratch moved to a `[ThreadStatic]` field, and the `Random` is now reseeded from a hash of the world seed and the column coordinates instead of advancing one shared sequence across the world.

That reseed also removes a dependency the old code already carried on dispatch order: two different columns always get two different seeds, so a column's result no longer depends on which column the dispatcher happened to hand out first.

## A pre-existing non-determinism, ruled out rather than assumed

Two full pregen runs of a 1257-column radius at a fixed seed both finished with zero worldgen exceptions and identical completion counts (1208 generated + 49 already loaded = 1257, both runs). But they did **not** produce identical per-column creature spawns, so that got chased down before trusting the fix.

A control pair of runs with the cap forced back to 1, serial dispatch, exactly what PreDone has always done, showed the same pattern. The cause sits outside this change: `entityTypeGroups` is a `Dictionary` keyed by `EntityProperties`, a reference type with no custom hash code, so its iteration order depends on each key's default object hash and is not stable across process launches. Every column loops that dictionary drawing random numbers in iteration order, so a different order gives a different sequence from an identical seed.

Pre-existing trait of the generator, confirmed at cap 1, out of scope here.

## The mod-safety fallback

| Axis | Catches | How |
| --- | --- | --- |

```
```

| --- | --- | --- | --- |

| --- | --- | --- | --- |

### Honest framing

## Set expectations for this branch specifically

PreDone is **0.2% of pregen time**. It hosts creature spawning and is the cheapest stage by a wide margin. This branch is a correctness prerequisite, not a throughput lever, and it will not show a gain on its own no matter how it is measured.

Its value is that it closes the last stage still pinned at cap 1, and that it carries the `dispatchClaimed` fix, which was root-caused on this branch and matters to all four.

## Verification

Three cases, end to end, at `MaxWorldgenThreads=4`, with a real minimal test mod built for this (own assembly, references `0Harmony`, dropped into `Mods/` for these tests only):

| Case | Expected | Result |
| --- | --- | --- |
| Mod registers its own PreDone handler | fall back, name the assembly | warned naming `SafeguardTestMod`, cap 1 |
| Mod Harmony-prefixes `GenCreatures.OnChunkColumnGen` | fall back, name mod + kind + method | warned naming `safeguardtestmod`, prefix, `GenCreatures.OnChunkColumnGen` |

Since PreDone hosts one generator, that third case is the entire second axis for this stage in practice.

All three reached RunGame with zero real crashes and the same 2 known false-positive log lines each (a class literally named `ErrorReporter`, and a line reading "no errors"). Plus 3 repeated boot trials: 0 stuck-column lines, 0 watchdog throws.

## Merge ordering

Independent of the other three cap-raise branches; merges cleanly into `indev` today. The shared `dispatchClaimed` and safeguard blocks are identical across all four, so the rest resolve trivially against whichever lands first.

/cc @tehtelev

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,16 @@ bootstrap: ## Download, decompile, and apply patches
 
 build: ## Build Release (runs bootstrap if working tree is missing)
 	@if [ ! -f VintagestoryApi/VintagestoryAPI.csproj ]; then $(MAKE) bootstrap; fi
-	dotnet build VintageStory.slnx -c $(CONFIGURATION) -p:EmbedPatchedFiles=$(EMBED_PATCHED_FILES)
+	dotnet build VintageStory.slnx -c $(CONFIGURATION)
+# Second pass, only if embedding is on: StratumServer's EmbeddedResource list
+# points at sibling projects' bin output by raw path, not a ProjectReference,
+# so on a build where those outputs do not exist yet (a fresh bootstrap, or
+# after clean/refresh) the embed can race the projects that produce them.
+# The pass above guarantees every output exists before this one tries to
+# embed it.
+ifeq ($(EMBED_PATCHED_FILES),true)
+	dotnet build VintageStory.slnx -c $(CONFIGURATION) -p:EmbedPatchedFiles=true
+endif
 
 smoke: build ## Build and boot-test the server
 	bash scripts/smoke-test.sh
--- a/patches/VSEssentials/Systems/WorldGen/Standard/ChunkGen/93.GenCreatures.cs.patch
+++ b/patches/VSEssentials/Systems/WorldGen/Standard/ChunkGen/93.GenCreatures.cs.patch
@@ -1,0 +1,252 @@
+diff --git a/VSEssentials/Systems/WorldGen/Standard/ChunkGen/93.GenCreatures.cs b/VSEssentials/Systems/WorldGen/Standard/ChunkGen/93.GenCreatures.cs
+index 647310c..8df05ba 100644
+--- a/VSEssentials/Systems/WorldGen/Standard/ChunkGen/93.GenCreatures.cs
++++ b/VSEssentials/Systems/WorldGen/Standard/ChunkGen/93.GenCreatures.cs
+@@ -20,35 +20,42 @@ namespace Vintagestory.ServerMods
+     }
+ 
+     public class GenCreatures : ModStdWorldGen
+     {
+         protected ICoreServerAPI api;
+-        protected Random rnd;
+         protected int worldheight;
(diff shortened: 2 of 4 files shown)


tehtelev (approved):
PRs 186, 187, and 190 were tested together with various numbers of threads. No critical errors were found related to this PR.

[server-main.log](https://github.com/user-attachments/files/30384985/server-main.log)

Zaldaryon:
No code change here, thanks for the approval. Flagging two things from the other reviews that affect this one, since you tested it stacked on #186+#187.

**Pick up #186 before merging this.** The warning you flagged there turned out to be a data-corruption bug that is live on `indev` today: a `[ThreadStatic] static` remap map shared across schematic instances ends up clearing a different schematic's permanent `BlockCodes`. It needs no concurrency to trigger and it explains three of your four reports across these PRs. Fixed in `aa46ef6`.

**The performance section in this body has been restated**, because two claims in it were wrong:

- The stage shares came from a radius-20 profile. At your radius-100 scale TerrainLate is 29.9% and Vegetation 27.7% while Terrain is 17.9%, not the 42.7% I originally quoted.
- Your Discord point about neighbour parity checks out, and I instrumented it. `ensurePrettyNeighbourhood` returns true immediately at or below TerrainLate and otherwise requires all 8 neighbours at the same pass. Measured across every run: Terrain and TerrainLate block exactly 0 times, the gated stages block heavily. So the honest framing is that the value is concentrated in the two ungated stages, and caps on gated stages are latent capability rather than throughput.

I also measured your queue hypothesis and it did not hold: radius 60, 4 interleaved samples per config, queue=2000 median 170.5s against queue=8000 median 170.0s, with 34-39s spreads. The exhaustion warnings do disappear, they just do not buy throughput. Worth knowing that every earlier radius-20 measurement in this initiative ran 1,681 columns against a 2,000 slot queue, so none of them could ever have stressed it, which is why I never saw what you saw.

PR #191, opened 2026-07-25
Let TerrainFeatures run several columns at once, and fix the two things that blocked it

Zaldaryon:
> [!NOTE]
> **Rebuilt on current `indev` (2026-07-27).** Four of the five commits this branch used to carry are already merged (via #186 and #187), which is why it showed 44 conflict markers against `indev`. Rather than resolve conflicts between commits and themselves, the branch was reset onto `indev` and now contains **one commit** with only what is genuinely new. It also sits on top of #199 and #188, both merged.

TerrainFeatures is the last worldgen pass whose cap is still 1 where raising it can bind. Getting there needed two other things fixed first, both found by actually running it.

## 1. The cap

Raised to the worker-thread count, with `EnforceStageConcurrencySafety` extended to the pass, matching every other raised cap.

## 2. Queue size 2000 → 8000

**TerrainFeatures is the first neighbour-gated pass to get a raised cap, and that is what makes this necessary.** `ensurePrettyNeighbourhood` returns `true` immediately for Terrain and TerrainLate, so they never enqueue anything. From TerrainFeatures up it requires all 8 neighbours at the same pass and **enqueues the laggards**. Running several of its columns at once multiplies those nudges.

Radius 100, seed `1027995113`, 3 worldgen threads:

| queue | result |
|---|---|
| 2000 | **3091** `Indexed Fifo Queue overflow` throws, frozen at 7.8% |
| 8000 | zero overflows |

The claim release from #199 works correctly throughout — no column is ever leaked — but `ScheduleReadyTasks` abandons its **whole dispatch pass** on any exception, so what #199 turned from a permanent wedge became a livelock instead. Raising the queue is what removes the cause.

## 3. Locking `ChunkDataLayer`'s writers

With the queue raised the run got further, and hit a different failure: **15 worldgen `NullReferenceException`s, 8 more on the read side, and a fatal one in `SaveDirtyUnloadedChunks` that killed the server two minutes in.**

**This one is vanilla's bug, not the fork's.** `MakeSpaceInPalette` is byte-identical to the decompiled original. It reassigns `palette`, allocates `dataBits[bitsize]`, then does a non-atomic `bitsize++`. Two threads growing together advance `bitsize` twice for one allocated plane, so a later grow either indexes `dataBits[15]` on a fixed-length `int[15][]`, or dereferences a plane nobody allocated — `IndexOutOfRangeException` and `NullReferenceException` respectively, both observed.

Vanilla's own guard cannot help:

```csharp
lock (palette)                      // locks the field...
{
    if (num == palette.Length) num = MakeSpaceInPalette();   // ...which this reassigns
    palette[num] = value;
}
```

A second thread entering after the reassignment locks a **different object**, and both critical sections run at once. `SetUnsafe` takes no lock at all — and it is the one `DiscDepositGenerator.GenDeposit` reaches.

Fixed with a dedicated per-layer lock held across the whole resolve-or-grow-then-write sequence in `Set`, `SetUnsafe`, `SetZero` and `SetBulk`, with the misleading `lock (palette)` removed. Locking rather than changing the structure because `palette` and `dataBits` **are** the on-disk and wire format — `CompressUsing`/`Decompress` serialise them directly and clients decode that same layout.

## Verification

Radius 100, seed `1027995113`, queue 8000, 3 worldgen threads, single variable against the crashing run:

| | without the lock | with it |
|---|---|---|
| result | **fatal at ~2 min** | completes, 31417/31417 columns |
| `NullReference` / `IndexOutOfRange` | 23 | **0** |
| `Server Fatal` | 1 | **0** |
| queue overflow | 0 | 0 |
| error/exception lines | 35 | 2 (the usual acceptable pair) |

Zero stuck columns and zero watchdog throws.

## On the lock's granularity

Per-block locking on a worldgen hot path is the obvious objection, so the alternatives are recorded in the code comment rather than left for someone to re-propose from theory:

1. **A per-chunk lock held across a whole generator call**, acquiring every touched chunk in index order to avoid deadlock. This is what C2ME does for Minecraft (`AsyncCombinedLock`/FlowSched). Cuts acquisitions by orders of magnitude.
2. **Single-writer-per-chunk confinement**, needing no lock at all. This is what vanilla Minecraft assumes — `PalettedContainer` does not synchronise, it uses `ThreadingDetector` to **deliberately crash** when two threads touch one container. Paper's Moonrise fork deletes even that check once ownership is provable.

Both are better in the abstract. Neither was adopted because both push correctness onto every caller, and a caller that forgets is exactly how this bug class keeps recurring here: `ServerMapChunk.NewBlockEntities` was locked for this same reason while the block write two lines above it in `BlockAccessorWorldGen.SetSolidBlock` was not — and that is one of the paths that crashed.

**The cost is not established.** 8m22s with the lock against 7m34s without, on single runs — about +10%, but that sits inside this machine's 6m06s–7m42s spread across comparable configurations the same day. Treat it as an unresolved upper bound, not a measured cost; interleaved sampling is owed before claiming either way.

## Disclosure

This branch produces `Tried to get block outside generating chunks!` notifications (2813 in the verifying run). Same class already disclosed on #188: every one comes from vanilla code reading across chunk boundaries, all are **reads**, and there are zero out-of-range **writes** in any run. `GetBlockId` returns air there, so the visible effect is a skipped or displaced loose rock. Not root-caused further.

## Upstream

The palette race is reachable in plain Vintage Story by anyone raising `MaxWorldgenThreads` — same class as `VintageStory-Issues#5143`. Worth reporting; not filed here.

### DIFF
--- a/patches/VintagestoryLib/Vintagestory.Common/ChunkDataLayer.cs.patch
+++ b/patches/VintagestoryLib/Vintagestory.Common/ChunkDataLayer.cs.patch
@@ -1,8 +1,81 @@
 diff --git a/VintagestoryLib/Vintagestory.Common/ChunkDataLayer.cs b/VintagestoryLib/Vintagestory.Common/ChunkDataLayer.cs
-index 4d0bf91..e9e511f 100644
+index 4d0bf91..caad0da 100644
 --- a/VintagestoryLib/Vintagestory.Common/ChunkDataLayer.cs
 +++ b/VintagestoryLib/Vintagestory.Common/ChunkDataLayer.cs
-@@ -56,103 +56,81 @@ public class ChunkDataLayer
+@@ -25,10 +25,72 @@ public class ChunkDataLayer
+ 
+ 	protected int bitsize;
+ 
+ 	public FastRWLock readWriteLock;
+ 
++	// Stratum: serialises every writer that can grow the palette or flip plane bits.
++	//
++	// The palette grows by MakeSpaceInPalette, which reassigns the palette field,
++	// claims dataBits[bitsize], and then does a non-atomic bitsize++. Vanilla guards
++	// the Set path with `lock (palette)`, which cannot work: the lock statement
++	// evaluates its argument once at entry, so once MakeSpaceInPalette has swapped
++	// the field, the next thread through locks the NEW array, a different monitor,
++	// and both critical sections run at once. SetUnsafe takes no lock at all.
++	//
++	// Two threads growing together make bitsize advance twice for one allocated
++	// plane. dataBits is a fixed int[15][], so the symptom is either
++	// IndexOutOfRangeException on dataBits[15] or NullReferenceException on a plane
++	// that was skipped, depending on the interleaving. Both were observed once
++	// TerrainFeatures ran several columns at once.
++	//
++	// Vanilla never exposed this because it never ran two threads into one
++	// ChunkDataLayer; the code here is byte-identical to vanilla's, so this is an
++	// upstream latent bug, not something the fork introduced. Locking rather than
++	// swapping in a concurrent structure because palette and dataBits ARE the
++	// on-disk and wire format (see CompressUsing/Decompress), so their shape cannot
++	// change without breaking saves and clients.
++	//
++	// On the granularity, since the obvious objection is that this locks per block
++	// written on a worldgen hot path. Two coarser designs were considered and are
++	// recorded here so they are not re-proposed from theory alone:
++	//
++	//   1. A per-chunk lock taken by callers and held across a whole generator or
++	//      structure-placement call, acquiring every touched chunk's lock up front
++	//      in index order to avoid deadlock. This is what C2ME does for Minecraft
++	//      (AsyncCombinedLock/FlowSched), and it cuts lock acquisitions by orders of
++	//      magnitude rather than shaving the per-acquisition cost.
++	//   2. Single-writer-per-chunk confinement, so no lock is needed at all. This is
++	//      what vanilla Minecraft assumes — its PalettedContainer does not
++	//      synchronise, it uses ThreadingDetector to deliberately crash when two
++	//      threads touch one container — and Paper's Moonrise fork deletes even that
++	//      check once ownership is provable.
++	//
++	// Both are better in the abstract. Neither was adopted, and the honest reason is
++	// structural rather than performance, because the performance case is weaker than
++	// it first looked. Measured at radius 100, same seed and configuration, single
++	// runs: 8m22s with this lock against 7m34s without, so about +10%. Mid-run
++	// progress looked identical (40.4% against 40.8% at three minutes), which is what
++	// made an earlier note here claim the cost was nil — it is not. The figure sits
++	// inside this machine's run-to-run spread, which ranged 6m06s to 7m42s across
++	// comparable configurations the same day, so n=1 cannot separate a real 10% from
++	// noise; treat it as an unresolved upper bound, not a measured cost. Interleaved
++	// sampling is owed before anyone claims either way. Part of the cost is avoided
++	// already: bulk terrain goes through SetBulk, which takes the lock once per bulk
++	// operation rather than once per block, so the per-block callers are deposits and
++	// structures, a minority of total writes.
++	//
++	// Structural, and the part that actually decided it: options 1
++	// and 2 both push correctness onto every caller, and a caller that forgets is
++	// exactly how this bug class keeps recurring here — ServerMapChunk.NewBlockEntities
++	// was locked for precisely this reason while the block write two lines above it in
++	// BlockAccessorWorldGen.SetSolidBlock was not, which is one of the paths that
++	// crashed. Locking centrally protects present and future writers without that
++	// audit. Revisit if profiling ever shows this lock hot, or if the worldgen
++	// scheduler grows a real per-chunk ownership model, at which point option 2 makes
++	// this deletable.
++	private readonly object stratumWriteLock = new object();
++
+ 	public Func<int, int> Get;
+ 
+ 	protected int[] dataBit0;
+ 
+ 	protected int[] dataBit1;
+@@ -56,103 +118,81 @@ public class ChunkDataLayer
  		Get = GetFromBits0;
  		readWriteLock = new FastRWLock(chunkDataPool);
  		pool = chunkDataPool;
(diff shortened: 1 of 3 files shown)


tehtelev (changes requested):
PRs 186, 187, and 191 were tested together with varying numbers of threads. No critical errors were found.
There is one new warning that I hadn't seen before. Something like this

`Could not import block entity data for schematic at 513735, 88, 510479. There is already game:block crate/11562, expected microblock. Possibly overlapping ruins, or is this schematic from an old game version?`

[server-main.log](https://github.com/user-attachments/files/30385136/server-main.log)

Zaldaryon:
The `already crate, expected microblock` warning you saw here has the same cause as the `block ID not found` warning on #186: a `[ThreadStatic] static` remap map that shares one slot across every schematic on a thread and ends up clearing a different schematic's permanent `BlockCodes`. Wrong block placed from a corrupted map, so the block-entity import finds a crate where it expected a microblock. Fixed on #186 in `aa46ef6`.

**I have not marked that as resolved here, on purpose.** The causal chain is sound but it is an inference: I have not observed the warning stop, and a smoke test cannot reproduce it. Closing it on reasoning alone is exactly the kind of thing that comes back. A retest of this branch on top of the fixed #186 is what would settle it.

Nothing else changed on this PR beyond the shared performance section in the body, which I have restated. Two corrections there worth your attention:

- The stage shares I originally quoted came from a radius-20 profile and were wrong at your scale. At radius 100 TerrainLate is 29.9% and Vegetation 27.7% while Terrain is 17.9%, not the 42.7% I claimed.
- Your neighbour-parity point from Discord checks out, and I instrumented it. `ensurePrettyNeighbourhood` returns true immediately at or below TerrainLate and otherwise needs all 8 neighbours at the same pass, so **TerrainFeatures is gated and Terrain and TerrainLate are not**. Measured: Terrain and TerrainLate block exactly 0 times per run, the gated stages block heavily. That explains why this branch and #190 both measure flat far better than my stage-cost argument did, and it means this cap is largely latent capability until the gate constraint moves.

I also tested your queue hypothesis, and it did not hold: radius 60, 4 interleaved samples per config, queue=2000 median 170.5s vs queue=8000 median 170.0s against 34-39s spreads. Exhaustion warnings do vanish (2 of 4 runs to 0 of 4) but buy no throughput. Full numbers are in the PR body.

Zaldaryon:
Rebuilt on current `indev` now that #190 and #189 have merged, so this shows no conflicts for the re-review.

Two things changed beyond the rebase:

- The shared dispatcher infrastructure now comes from `indev` rather than being duplicated here, so this branch adds only TerrainFeatures' own cap raise and safeguard call.
- It merges its two prerequisites rather than restating them, so the branch is exactly the stack you test: #186 (which now carries the `BlockCodes` fix as well as the `Unpack` lock) and #187 (which now carries the `GenPonds` guard).

That second point matters for your warning here specifically: the `already crate, expected microblock` line should be gone, because its cause was the `BlockCodes` destruction fixed on #186, and that fix is now in this branch. **I still have not marked it resolved**, since I have reasoned that rather than observed it, and this is exactly the branch where a retest would settle it.

Verified after the rebuild: build 0 errors, `smoke-test.sh` reaches RunGame, exactly the 2 known false-positive log lines.

Zaldaryon:
Rebuilt on current `indev` after the `BlockCodes` corruption fix merged as #195, so this no longer carries it and shows no conflicts.

The branch now merges its two remaining prerequisites, #186 (reduced to the `Unpack` lock and the SQLite read locks) and #187, and adds only TerrainFeatures' own cap raise and safeguard call.

Your `already crate, expected microblock` warning here should be gone: its cause was the `BlockCodes` destruction, which is now fixed on `indev` and therefore in this branch. I have **not** marked it resolved, since I reasoned that rather than observed it, and this is the branch where a retest would settle it.

Zaldaryon:
**This branch now has a hard prerequisite: #196.**

I merged every open worldgen PR into one throwaway branch and reran the radius-100 reproduction to check what the end state actually looks like. The `BlockCodes` corruption fixed in #195 was gone, but a second race surfaced underneath it, which that corruption had been hiding by aborting schematic placement before the code was reached:

```
InvalidOperationException: Operations that change non-concurrent collections must have exclusive access.
  at HashSet`1.AddIfNotPresent
  at BlockSchematic.PlaceEntitiesAndBlockEntities ... in pass TerrainFeatures
```

`ServerMapChunk.NewBlockEntities` is a plain `HashSet<BlockPos>` written by worldgen threads, drained by the chunk thread, and enumerated by the save path, with nothing synchronising any of it. Structures cross chunk boundaries, so two columns in a structure-placing pass reach the same map chunk's set. **Raising TerrainFeatures' cap, which is what this PR does, is what makes it reachable.**

Fixed in #196, verified on the same reproduction: with every open PR merged, worldgen exceptions go from 1 to 0.

So the ordering for this branch is now #195 (merged), then #196, then this. I have not folded #196 into here, keeping to one PR at a time.

Zaldaryon:
**Holding this one too, with evidence.** The other worldgen PRs are now merged (#195, #196, #186, #187, #194) and `indev` validates clean at radius 100 with your seed: zero worldgen exceptions, zero beam-mapping warnings, zero overlapping-ruin warnings.

This branch does not. Two radius-100 runs of everything merged, same seed:

| Run | worldgen exceptions | overlapping-ruin warnings |
| --- | --- | --- |
| 1 | 0 | 4 |
| 2 | **2** | **6** |

The first run is why I nearly declared this clean. The second shows it was luck. The exceptions are a class I had not seen before:

```
IndexOutOfRangeException at ChunkDataLayer.Set        <- via BlockSchematicStructure.PlaceReplacingBlocks
IndexOutOfRangeException at ChunkDataLayer.SetUnsafe  <- via DiscDepositGenerator.GenDeposit
```

Both in pass TerrainFeatures. That is concurrent writes into the same chunk's packed block storage, which is a deeper problem than the `NewBlockEntities` HashSet that #196 fixed: #196 guarded a set of positions, this is the block data itself.

Worth being explicit about the pattern, since it has now happened twice: **fixing one race exposes the next.** The `BlockCodes` corruption hid the `NewBlockEntities` race by aborting placement before it was reached, and fixing that exposed these `ChunkDataLayer` ones. Testing this branch stacked on the others is what let each one hide the next, which is the argument for merging one at a time that you made.

Also relevant to the chiseled-block question: the `expected microblock` warnings are **gone on merged `indev`** but come back on this branch, 4 in one run and 6 in the other. So they track this cap raise, not the `BlockCodes` bug that #195 fixed.

I would keep this closed until the `ChunkDataLayer` writes are understood. TerrainFeatures is 16% of stage time, so it is not worth shipping a race for.

PR #195, opened 2026-07-27
Stop the schematic remap map from clearing another schematic's BlockCodes

Zaldaryon:
Fixes the `KeyNotFoundException` in `BlockSchematic.PlaceDecors` and the `block id N not found in block registry` warnings reported against **`v1.22.5-stratum.1`**, the current stable release.

One file. No behaviour change beyond the fix, no new API, none of the other worldgen work in flight, so it can ship as a patch release on its own.

## The bug

`BlockCodesTmpForRemap` used to be a **per-instance** field. It was converted to `[ThreadStatic] private static`, which is one slot per thread shared by **every** `BlockSchematicStructure` on that thread, and the field was made to hold either the scratch dictionary or `BlockCodes` itself depending on which branch ran:

```csharp
if (replaceBlocks == null)
{
    stratumBlockCodesTmpForRemap = BlockCodes;   // aliases THIS schematic's permanent map
    return;
}
if (stratumBlockCodesTmpForRemap == null || stratumBlockCodesTmpForRemap == BlockCodes) stratumBlockCodesTmpForRemap = new Dictionary<int, AssetLocation>();
else stratumBlockCodesTmpForRemap.Clear();      // clears whatever it points at
```

The identity guard only compares against **this** instance's `BlockCodes`, so it cannot tell the slot is aliasing a *different* schematic's map:

1. Schematic **A** places with `replaceBlocks == null`, and the slot now aliases `A.BlockCodes`.
2. Schematic **B** places on the same thread with `replaceBlocks != null`, the slot is non-null and `!= B.BlockCodes`, so it falls through and **clears `A.BlockCodes`**.

`BlockCodes` is the schematic's permanent id to `AssetLocation` map, loaded once and reused for the process lifetime, so A stays broken from then on. `BlockSchematic.PlaceOneDecor` indexes it bare (`BlockCodes[id]`), which is the `KeyNotFoundException`, and the wrong blocks placed before it throws are **written into the world and persist**.

**No concurrency is required.** `WorldGenStructure.resolvedRockTypeRemaps` starts `null` and is only populated when a structure configures a rock-type remap group, so structures without remaps supply step 1 and structures with them supply step 2, and a normal world has both. The crash lands in the **TerrainFeatures** pass, whose concurrency cap is 1 in the released build, which confirms this independently of any cap raise.

## The fix

Remove the shared publication point. `resolveReplaceRemapsForBlockEntities` returns the map, both callers hold it in a local and pass that local to `PlaceEntitiesAndBlockEntities`, and the `replaceBlocks == null` path returns `BlockCodes` untouched, which is what vanilla `BlockSchematic` already does at the equivalent call. The thread-static stays purely as reuse scratch for the remap case, where it can now only ever hold its own dictionary, so clearing and reusing it is safe.

## Verified by reproduction, not by reasoning

Radius 100 pregen, 3 worldgen threads, seed `1027995113`, identical settings both sides:

| | `indev` (before) | this branch (after) |
| --- | --- | --- |
| `KeyNotFoundException` in `PlaceDecors` | **2** | **0** |
| worldgen exceptions | **1** | **0** |
| pregen wall clock | 6m45s | 6m49s |

The reproduced stack trace matches the reported one line for line, down to `GenStructures.OnChunkColumnGenPostPass` → `TryGenerateAtSurface` → `PlaceRespectingBlockLayers` → `PlaceDecors`. Costs nothing measurable.

### DIFF
--- a/patches/VSEssentials/Systems/WorldGen/Standard/Datastructures/BlockSchematicStructure.cs.patch
+++ b/patches/VSEssentials/Systems/WorldGen/Standard/Datastructures/BlockSchematicStructure.cs.patch
@@ -1,51 +1,77 @@
 diff --git a/VSEssentials/Systems/WorldGen/Standard/Datastructures/BlockSchematicStructure.cs b/VSEssentials/Systems/WorldGen/Standard/Datastructures/BlockSchematicStructure.cs
-index 7573f32..03c54ee 100644
+index 7573f32..27a7943 100644
 --- a/VSEssentials/Systems/WorldGen/Standard/Datastructures/BlockSchematicStructure.cs
 +++ b/VSEssentials/Systems/WorldGen/Standard/Datastructures/BlockSchematicStructure.cs
-@@ -12,11 +12,19 @@ using Vintagestory.ServerMods.NoObf;
+@@ -12,11 +12,30 @@ using Vintagestory.ServerMods.NoObf;
  
  namespace Vintagestory.ServerMods
  {
      public class BlockSchematicStructure : BlockSchematic
      {
 -        public Dictionary<int, AssetLocation> BlockCodesTmpForRemap = new();
-+        // Stratum: was a shared instance field, written by resolveReplaceRemapsForBlockEntities
-+        // and read a few lines later by PlaceEntitiesAndBlockEntities, both within the same
-+        // placement call. Different columns placing the same cached schematic concurrently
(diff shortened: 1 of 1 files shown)


tehtelev:
I tested pre-generation at a radius of 100 with 6 threads, as well as at a radius of 150 with 3 threads—0 errors.
Good work. You can take a break for now.

PR #206, opened 2026-07-28
Fixing bugs with lighting and foliage 

### LINKED ISSUE #198: Lighting issue with leaf blocks inner planes
### Stratum version

v1.22.5-stratum.1

### Base game version

1.22.5

### OS and .NET runtime

Windows 11

### Steps to reproduce

Created a brand new world on 1.22.5, no mods. Joined world and noticed black / dark leaf planes.

### Expected behavior

Lighting should absorb into blocks with low-absorption

### Actual behavior

Lighting does not absorb into low-absorption blocks at all.

Image
Image
Image
Image

### Server log

```text
Not needed, nothing included.
```

### Checks

- [x] I confirmed this does not happen on a stock vanilla server.
- [x] I am running the latest Stratum release.

tehtelev:
## Summary

The following improvements have been made:
- The function for calculating effective absorption has been refined and now also accounts for absorption by low-absorption blocks such as foliage, etc.
- The calculation formula now also accounts for a block’s absorption, even if the ray was able to pass through it. It is guaranteed that half of the light passes through exactly, while the other half is reduced in proportion to the absorption. For example, if a ray hits the edge of a slab, light with an intensity of 20 will pass through the slab tangentially with an intensity of 10, since the slab’s absorption is greater than 32 (we assume the remaining 10 was absorbed by the slab). If the slab is made of glass, 20 units of light will always pass through.
- Patches have been implemented that allow block light and sunlight to be rendered into blocks with solid faces. This was done to ensure that foliage is rendered correctly. This is because foliage textures can, for some reason, draw brightness from neighboring blocks, which resulted in unpredictable coloring of the leaves.

**Before**

**After**

## Type

- [x] Bug fix
- [ ] Performance
- [ ] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

## Performance numbers

No changes

## Related issues

Fixes #198

### DIFF
--- a/patches/VintagestoryLib/Vintagestory.Common/ChunkIlluminator.cs.patch
+++ b/patches/VintagestoryLib/Vintagestory.Common/ChunkIlluminator.cs.patch
@@ -1,8 +1,8 @@
 diff --git a/VintagestoryLib/Vintagestory.Common/ChunkIlluminator.cs b/VintagestoryLib/Vintagestory.Common/ChunkIlluminator.cs
-index e2ca303..ae4762d 100644
+index e2ca303..da7c67d 100644
 --- a/VintagestoryLib/Vintagestory.Common/ChunkIlluminator.cs
 +++ b/VintagestoryLib/Vintagestory.Common/ChunkIlluminator.cs
-@@ -1,1110 +1,1592 @@
+@@ -1,1110 +1,1641 @@
  using System;
  using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


Zaldaryon:
Available in v1.22.5-stratum.1-indev.3: https://github.com/StratumServer/Stratum/releases/tag/v1.22.5-stratum.1-indev.3

Issue #224, opened 2026-07-29
Generation errors in the ruins schematics


tehtelev:
### Stratum version

1.22.5-stratum.1 (latest version of indev branch)

### Base game version

1.22.5

### OS and .NET runtime

Windows 10 Корпоративная, 21H1, NET 10

### Steps to reproduce

1. Build the server using the latest version of the indev branch.
2. Set the number of generation threads to 2 or more.
3. Run pre-generation for 50–100 chunks.

### Expected behavior

During world generation or pre-generation, there should be no structure generation errors, regardless of the number of threads.

### Actual behavior

1 thread (no errors).

3 threads

[server-main.log](https://github.com/user-attachments/files/30505622/server-main.log)
[server-worldgen.log](https://github.com/user-attachments/files/30505657/server-worldgen.log)

6 threads
[server-main.log](https://github.com/user-attachments/files/30505925/server-main.log)

### Server log

```text

```

### Checks

- [x] I confirmed this does not happen on a stock vanilla server.
- [x] I am running the latest Stratum release.

Zaldaryon:
Fixed in https://github.com/StratumServer/Stratum/pull/225

PR #225, opened 2026-07-29
Hold the neighbourhood resident while a worldgen stage claim lasts

Zaldaryon:
## The bug

`ensurePrettyNeighbourhood` checks that a column's 8 neighbours are resident before it grants a stage claim, and `EnsureMinimumWorldgenPassAt` backs that check: it passes a neighbour only when that neighbour is a live request or a loaded map chunk at pass Done. Residency holds at the moment of the check. Nothing holds it after.

The unload pass then drops one of those columns while the generator reads across the boundary. `GetGeneratingChunk` stops answering once the column leaves the request queue. The loaded chunks fallback stops answering once it leaves memory. `BlockAccessorWorldGen.GetBlockId` then returns 0, and generators take that 0 for an air block, so a failed read does not stop them, it hands them a wrong answer.

`GenPonds` shows it first. Its flood fill decides pond extent by walking air across the 3x3 column neighbourhood, so a neighbour it cannot resolve reads as open space.

## The fix, in two parts

Pin the 3x3 neighbourhood for the length of the claim. `TryClaimStage` takes the pin before the gate runs rather than after it passes, since a hold that starts after the check leaves open the gap the check exists to close. Every path that releases the claim releases the pin: both worker paths, the dispatch failure path, and the inline runner.

Interlock that pin against the unload pass. The pass decides and removes in two steps: it reads the pin, then it saves the chunks and drops the column. A pin landing in that window goes unseen, and the gate runs afterwards and sees a column about to disappear. The pass now sets `stratumDropping` before it touches anything and reads the pin once more, and the gate treats a flagged column as absent. A pin that lands first makes the pass back out. A drop that commits first makes the gate refuse the claim. Neither ordering grants a claim over a column under removal.

The second part carries its weight. Pinning on its own leaves the race open, and the numbers show it.

## Measurements

Radius 100, seed 1027995113, 3 worldgen threads, five runs per condition, counting reads that fell outside the generating set:

| condition | runs | median |
|---|---|---|
| no pin | 213, 382, 474, 622, 1086 | 474 |
| pin only | 35, 87, 125, 249, 677 | 125 |
| pin and flag | 0, 0, 0, 0, 0 | 0 |
| serial worldgen | 0 | 0 |

Pin and flag against no pin: complete separation, p about 0.008 on Mann-Whitney. Pin on its own against no pin misses significance at n = 5 per side, and every event that survived pinning alone landed inside the 3x3, at a column holding a pin when the read failed. That pointed at the window between the decision and the removal.

Serial worldgen produces zero here, at radius 40 and at radius 100, which makes this a concurrency regression rather than a limit of the base game.

Two more runs from a tree rebuilt out of these patches produced zero reads and zero worldgen errors, at 5m17s and 5m01s against 5m13s to 6m36s for the control. Holding the columns costs nothing I can measure. Active claims times nine bounds the pinned set.

## Notes for review

The release sites carry the risk. A leaked pin starves the unload pass, a missing one reopens the race, so every path that writes `dispatchClaimed` back to 0 also unpins, and the flag comes down even on the throw inside the removal. That symmetry follows the lesson of #199.

Single runs at this radius span 213 to 2786. Anything measured here needs repeats. An early version of this patch looked finished on one run and was not.

## On #224

@tehtelev this may well be your issue, and I would rather you check than take my word for it. #224 fails on `GetMapChunkAtBlockPos` returning null into `BlockSchematicStructure.PlaceRespectingBlockLayers`. That lookup resolves through the same two registries this patch keeps populated, `loadedMapChunks` and the request queue, so keeping the column resident should close it. Treat that as reasoning rather than measurement: your NRE never fired here, across more than ten pregen runs at radius 100 on seed 1027995113 with 3 and 6 worldgen threads, on Linux with .NET 10.0.301, all of them showing zero `error was thrown in pass TerrainFeatures` entries. The lighter symptom did show once, one `Could not import block entity data for schematic ... Possibly overlapping ruins` on the 6 thread run against 7 in your log. The trigger is thread scheduling, so Windows against Linux is my best guess for why it fires for you and not here.

If you have the time, two things would settle it. Does your repro throw on this branch, and does `Tried to get block outside generating chunks` show up in your logs on a run without it.

### DIFF
--- a/patches/VSEssentials/Systems/WorldGen/Standard/ChunkGen/1.GenTerra/GenTerraPostProcess.cs.patch
+++ b/patches/VSEssentials/Systems/WorldGen/Standard/ChunkGen/1.GenTerra/GenTerraPostProcess.cs.patch
@@ -1,8 +1,8 @@
 diff --git a/VSEssentials/Systems/WorldGen/Standard/ChunkGen/1.GenTerra/GenTerraPostProcess.cs b/VSEssentials/Systems/WorldGen/Standard/ChunkGen/1.GenTerra/GenTerraPostProcess.cs
-index 79c481d..691b108 100644
+index 79c481d..c7a1523 100644
 --- a/VSEssentials/Systems/WorldGen/Standard/ChunkGen/1.GenTerra/GenTerraPostProcess.cs
 +++ b/VSEssentials/Systems/WorldGen/Standard/ChunkGen/1.GenTerra/GenTerraPostProcess.cs
-@@ -13,12 +13,24 @@ namespace Vintagestory.ServerMods
+@@ -13,12 +13,25 @@ namespace Vintagestory.ServerMods
      public class GenTerraPostProcess : ModStdWorldGen
      {
          ICoreServerAPI api;
@@ -13,7 +13,8 @@ index 79c481d..691b108 100644
 +        // Stratum: every field below used to be a plain instance field, written
 +        // and read within the processing of a single column (OnChunkColumnGen ->
 +        // deletePotentialFloatingBlocks -> AddToChunkVisitedNodesIfSameChunk),
-+        // unsafe once TerrainFeatures' same-stage cap moves off 1. Matches the
++        // unsafe once TerrainFeatures' same-stage cap moved off 1, which it since has
++        // (see ServerSystemSupplyChunks.stratumStageMaxConcurrent). Matches the
 +        // InvalidOperationException: Collection was modified crash reported
 +        // against this class: two columns concurrently mutating the same shared
 +        // HashSet/List. [ThreadStatic] since GenTerraPostProcess is a singleton
@@ -29,7 +30,7 @@ index 79c481d..691b108 100644
          {
              return side == EnumAppSide.Server;
          }
-@@ -56,10 +68,11 @@ namespace Vintagestory.ServerMods
+@@ -56,10 +69,11 @@ namespace Vintagestory.ServerMods
              const int chunksize = GlobalConstants.ChunkSize;
              const int chunksizeSquared = chunksize * chunksize;
              int chunkY = seaLevel / chunksize;
@@ -41,7 +42,7 @@ index 79c481d..691b108 100644
              for (int cy = chunkY; cy < cyMax; cy++)
              {
                  IChunkBlocks chunkdata = chunks[cy].Data;
-@@ -106,28 +119,29 @@ namespace Vintagestory.ServerMods
+@@ -106,28 +120,29 @@ namespace Vintagestory.ServerMods
              }
          }
  
@@ -75,7 +76,7 @@ index 79c481d..691b108 100644
  
              int baseX = X - halfSize;
              int baseY = Y - halfSize;
-@@ -225,11 +239,11 @@ namespace Vintagestory.ServerMods
+@@ -225,11 +240,11 @@ namespace Vintagestory.ServerMods
              if (((nposZ ^ origZ) & chunkMask) != 0) return;
              if (((nposY ^ origY) & chunkMask) != 0) return;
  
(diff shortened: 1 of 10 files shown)


tehtelev (changes requested):
Error
`Could not import block entity data for schematic ... Possibly overlapping ruins`
is still appearing.

`Tried to move the block outside the generating chunks`
It no longer appears


3 threads
[server-main.log](https://github.com/user-attachments/files/30516096/server-main.log)

6 threads
[server-main.log](https://github.com/user-attachments/files/30516234/server-main.log)

Zaldaryon:
The residency race this PR targets is closed. Neither log shows `Tried to move the block outside the generating chunks` or `Tried to get block outside generating chunks`, and neither shows a `NullReferenceException` or an `error was thrown in pass TerrainFeatures` entry.

The overlapping-ruins warning is a real bug, but not this one. `BlockSchematic.PlaceEntitiesAndBlockEntities` (the code that logs it) is a detector, not the damage: by the time it runs, `GenStructures`/`WorldGenVillage` has already placed two structures on top of each other. The actual defect is in structure placement, exposed by 86f70af raising TerrainFeatures off same-stage cap 1:

- `WorldGenStructure.WouldOverlapAt` reads `region.GeneratedStructures` to decide if a spot is free, but registration only happens after the caller's `TryGenerate` returns (`GenStructures.cs`). Two columns can both pass the check against a list neither has registered into yet, and both place.
- `WorldGenStructure`/`WorldGenVillage` instances are shared, not per-thread (`scfg.Structures`, `vcfg.VillageTypes`), and `TryGenerate` writes shared mutable state (`rand`, `tmpPos`, `LastPlacedSchematicLocation`) that the caller reads back after the call returns. A second thread's placement can overwrite that state before the first thread reads it.
- `ServerMapRegion.AddGeneratedStructure` is copy-on-write with no lock: two concurrent registrations can race, and one silently disappears.

None of this is in the residency-pin patch (`ChunkColumnLoadRequest`/`ChunkServerThread`/`ServerMain`/`ServerSystemSupplyChunks`/`ServerSystemUnloadChunks`/`WorldAPI`). I've added a lock serializing the decide-place-register sequence in `GenStructures` (ruins and villages both, since they share the same region's overlap check) and made `ServerMapRegion`'s registration atomic, so a lost or racing write can't happen regardless of which generator causes it.

Verification: seed 1027995113, radius 100, 6 worldgen threads, same coordinates (chunks 15900,15900 to 16100,16100) that produced the exact warning in the research archive here pre-fix. With the lock: pregen complete in 14m 49s, 31417/31417 columns, zero `overlapping ruins` lines, zero NREs. Same seed and radius reproduced the warning at these exact coordinates on every prior run without the lock.

Doesn't settle #224 either way. Still no repro of the NRE here, across every run at this radius on this branch.

One honest caveat from this same verification run: a single `Tried to get block outside generating chunks!` did fire once, unrelated to autosave, at ~85% through the run. That's the residency-pin's own symptom, not this fix's, and it's not in anything this comment's patch touches (`ChunkColumnLoadRequest`/`ChunkServerThread`/`ServerMain`/`ServerSystemSupplyChunks`/`ServerSystemUnloadChunks`/`WorldAPI` are untouched here). No stack trace (didn't have `RuntimeEnv.DebugOutOfRangeBlockAccess` on), and the box this ran on was badly CPU-starved for the whole run (ticks regularly at 500-700ms against a 33ms target), which is a plausible enough cause on its own that I wouldn't read one occurrence as a real gap in the pin. Flagging it rather than burying it, since one clean run isn't proof either way.

Zaldaryon:
@tehtelev could you re-run your same two configs (3 and 6 threads) against `e5ab9c4` and check whether the overlapping-ruins warning is gone on your end too? Everything above is from my own environment; I'd rather you confirm than take my word for it, same as the original PR asked of you for the residency fix.

tehtelev:
Tested e5ab9c4.
It looks like everything is fine now.
If necessary, I can run some more load tests on current usage in dottrace. I can also check the allocations in dotMemory.


3 threads
[server-main.log](https://github.com/user-attachments/files/30533719/server-main.log)

6 threads
[server-main.log](https://github.com/user-attachments/files/30533890/server-main.log)

Zaldaryon:
> Tested [e5ab9c4](https://github.com/StratumServer/Stratum/commit/e5ab9c426221052ef72b94e77b1af5650705031b). It looks like everything is fine now. If necessary, I can run some more load tests on current usage in dottrace. I can also check the allocations in dotMemory.
> 
> 3 threads [server-main.log](https://github.com/user-attachments/files/30533719/server-main.log)
> 
> 6 threads [server-main.log](https://github.com/user-attachments/files/30533890/server-main.log)

Please do. I'll merge either way now.

Zaldaryon:
Available in v1.22.5-stratum.1-indev.3: https://github.com/StratumServer/Stratum/releases/tag/v1.22.5-stratum.1-indev.3

Issue #230, opened 2026-08-01
World gen errors with BetterRuins on creativebuilding (superflat) worlds


Pixnop:
### Stratum version

v1.22.6-stratum.1-indev.1

### Base game version

1.22.6

### OS and .NET runtime

Linux, .NET 10, container based on mcr.microsoft.com/dotnet/runtime:10.0

### Steps to reproduce

Generate a world once so the config exists, switch it to the creative playstyle, wipe the save, start again:

```bash
jq '.WorldConfig.PlayStyle="creativebuilding"
    | .WorldConfig.PlayStyleLangCode="creativebuilding"
    | .DefaultRoleCode="crplayer"' serverconfig.json > c && mv c serverconfig.json
rm -rf Saves
./StratumServer --dataPath=. --stratum-no-banner
```

Count with `grep -c 'error was thrown in pass'`. BetterRuins 0.6.3 is the only mod needed.

### Expected behavior

No worldgen errors. The stock 1.22.6 dedicated server generates the same world with the same mod and the same playstyle cleanly.

### Actual behavior

169 errors during generation. Fresh data path each run, world generated from scratch.

| Runtime | Playstyle | Mods | Errors |
|---|---|---|---|
| Stratum 1.22.6-indev.1 | creativebuilding | betterruins only | 169 |
| Stratum 1.22.6-indev.1 | creativebuilding | none | 0 |
| Stratum 1.22.6-indev.1 | creativebuilding | full 39 mod pack | 197 |
| Stratum 1.22.6-indev.1 | surviveandbuild | full pack minus watersheds | 0 |
| Stock 1.22.6 | creativebuilding | betterruins only | 0 |
| Stock 1.22.6 | creativebuilding | none | 0 |

The playstyle dependency is the interesting part: the same mod on surviveandbuild under Stratum is clean.

### Server log

```text
Exception System.NullReferenceException: Object reference not set to an instance of an object.
   at Vintagestory.ServerMods.BlockSchematicStructure.SatisfiesMinSpawnDistance(Int32 minSpawnDistance, BlockPos pos, BlockPos spawnPos) in VSEssentials/Systems/WorldGen/Standard/Datastructures/BlockSchematicStructure.cs:line 56
[...]

PR #232, opened 2026-08-02
Indev -> main

trevorftp:


Zaldaryon (approved):

Pixnop (approved):
Validated this release candidate locally against the official indev.2 artifact, which matches this branch tip.

Full StratumParity differential suite, vanilla 1.22.6 vs 1.22.6-stratum.1-indev.2: 17/17 on both flavors, zero outcome divergence (only the expected duration asymmetry on the absence-proving random-tick probe). Fresh-extract first boot is clean, and the #231 worldgen compat work is present in the shipped binaries.

One note that removes a worry rather than adding one: the BetterRuins NRE from #230 reproduces identically on the official vanilla 1.22.6 server (same stack, same error count, controlled run with the playstyle confirmed in the log), so it is a vanilla VSEssentials bug, not a Stratum regression, and not a blocker for this release. I will follow up on #230 accordingly.

PR #237, opened 2026-08-07
Add mutual invisibility for vanished staff

### LINKED ISSUE #213: /vanish modification
### Problem

/vanish is lovely being able to turn invisible within creative as a staff member yet keep all the great qualities of creative mode without having to go into spectator mode and be very limited. Problem is when several staff are near each other in /vanish, they can still see each other. This complicates recording because you don’t want several people flying around that could be invisible and ruining the recording of players that you’re trying to get. 

### Proposed change

Add a toggle option to where you can have staff that are in /vanish mode, also become invisible for you as a staff member 

### Alternatives considered

_No response_

### Scope

Zaldaryon:
## Summary

`/vanish hideothers on|off|toggle`, a per-viewer preference. When on, a staff member stops seeing other vanished staff, without touching what anyone else sees. Also `/vanish status`, prints your own vanish state and this preference, changes nothing.

```
/vanish                          unchanged, toggle own vanish
/vanish status                   new
/vanish hideothers on|off|toggle new
```

New config default, `Commands.VanishHideOtherVanishedDefault` (bool, default false), the starting value for a staff member who never touched the preference. Default keeps today's behavior, vanish holders still see each other unless someone opts in.

### Where it hooks and why

`ShouldHideEntityFromClient(entity, client)` already takes the viewer, not just the subject. The whole feature is one added branch there: non-staff still get an unconditional `true`, staff now return their own `HidesOtherVanished(viewer.PlayerUID)` instead of the old unconditional `false`. Nothing else calls this predicate directly, so no call site needed to change.

The tracking sweep in `PhysicsManager` won't despawn a hidden entity on its own at the range this issue is actually about. It has a hysteresis window, an already-tracked entity stays tracked as long as it's inside 1.21x the tracking radius, which is exactly two staff standing next to each other. So the toggle needs its own packet push or it silently does nothing at close range while appearing to work at long range. `RefreshVanishedVisibilityForViewer` is that push, the inverse of the existing `HideVanishedPlayerFromOthers`/`RevealPlayerToOthers`: those loop every client for one subject, this loops every vanished subject for one client.

`HandleNear` had its own inline copy of the old visibility rule (`IsVanished(...) || PlayerHasAccess(...)`), so it would have kept listing a staff member the viewer just chose to stop seeing. Updated it to match.

### Testing detail

No test project in this repo, same live-boot convention as always. Live test with one connected account (a second wasn't available this round): `/vanish`, `/vanish status`, `/vanish hideothers on`, status again to confirm the flip, `/near` with nothing to hide (no crash), `/vanish hideothers off`, then bare `/vanish` to confirm it still toggles vanish and not the preference. `stratum-audit.log` shows `vanish hideothers=on actor=Zaldaryon`, confirming the command path end to end.

What that doesn't cover: the actual mutual-invisibility effect needs two vanished staff who can see each other's screens, and `RefreshVanishedVisibilityForViewer`'s despawn push is exactly the part most likely to be wrong in a way single-account testing can't catch (see the hysteresis note above, wrong and it's an invisible bug at close range and a working one at long range). Worth a two-account pass before merge if that's easy on your end, happy to do it myself if a second account gets easier to arrange here.

@trevorftp @tehtelev

## Type

- [ ] Bug fix
- [ ] Performance
- [x] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

## Related issues

Fixes #213

### DIFF
--- a/sources/VintagestoryLib/Vintagestory.Server/CmdStratumStaffCommands.cs
+++ b/sources/VintagestoryLib/Vintagestory.Server/CmdStratumStaffCommands.cs
@@ -156,8 +156,10 @@ internal class CmdStratumStaffCommands
 		if (StratumCommandRegistration.ShouldRegister(commands.Vanish, "/vanish", "Commands.Vanish"))
 		{
 			server.api.commandapi.Create("vanish")
-				.WithDescription("Toggle staff vanish")
-				.WithArgs(parsers.OptionalWordRange("mode", "on", "off", "toggle"))
+				.WithDescription("Toggle staff vanish, or /vanish hideothers to hide other vanished staff from yourself")
+				.WithArgs(
+					parsers.OptionalWordRange("mode", "on", "off", "toggle", "hideothers", "status"),
+					parsers.OptionalWordRange("state", "on", "off", "toggle"))
 				.RequiresPrivilege(Privilege.chat)
 				.HandleWith(HandleVanish);
 		}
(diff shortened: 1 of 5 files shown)


Zaldaryon:
Added a follow-up commit (f383b31): vanished players' pins are now also hidden from the world map.

Stratum already had `PlayerMapDisclosureHook` (wired up by `StratumPlayerPrivacy` for the separate opt-in map-pin privacy feature), so this wires vanish state into it instead of adding something new. Same visibility rule as entity hiding and `/near`: non-staff never see a vanished player's pin, and staff who've opted into `/vanish hideothers` stop seeing it too. This is unconditional — it doesn't depend on the `PlayerPrivacy.Enabled` config, since vanish is meant to hide the player entirely, not just off the entity tracker.

Build green (`dotnet build VintageStory.slnx -c Release -p:EmbedPatchedFiles=true`), no errors.

PR #238, opened 2026-08-07
Add /restart and /clearitems commands

### LINKED ISSUE #219: Server restart command and clear ground items
### Problem

Players do not easily know when the server is restarting and complain about it shutting down even though it was shut down on purpose. /announce helps but can get buried in chat and takes time to do when you have a lot of players to answer to what’s happening in a timely manner. 

### Proposed change

Having it be more automated makes it easier. When you type for example /restart 5 it displays at the center of the players screen “restarting server in 5 minutes” and also put that into global chat. It’ll count down in chat every minute that passes so “server restarts in 2 minutes” then it says 1 minute, then 45 seconds, 30 seconds, 15 seconds, 5 seconds. When that happens it’ll also say in chat “PICK UP ALL GROUND ITEMS THEY WILL BE CLEARED IN 30 SECONDS” then a clear ground items activation is done before server restart. A clear ground items command would be good as well then when activated it also says in global chat the “PICK UP ALL GROUND ITEMS THEY WILL BE CLEARED IN 30 SECONDS”

### Alternatives considered

_No response_

### Scope

Zaldaryon:
## Summary

Adds two staff commands for server lifecycle management.

**`/restart <1-60>`** schedules a server restart with countdown warnings broadcast to all players at configurable intervals (default: 5m, 2m, 1m, 45s, 30s, 15s, 5s). Ground items are cleared before the stop sequence when `ClearGroundItemsBeforeStop` is true. The countdown is cancelable via `/restart cancel`.

**`/clearitems [now|cancel]`** triggers a manual ground-item cleanup. Without arguments, it broadcasts a 30-second pickup warning first, cancelable via `/clearitems cancel`. With `now`, items are removed at once.

Both commands use the existing `StratumCommandAccess` permission system (`stratum.restart`, `stratum.clearitems`). Configuration lives in `Performance.Restart` within `stratum-performance.json`:

```json
{
  "CountdownAnnouncementsSeconds": [300, 120, 60, 45, 30, 15, 5],
  "ClearGroundItemsBeforeStop": true,
  "ClearGroundItemsLeadSeconds": 30,
  "CountdownMessage": "Server restarting in {0}.",
  "ClearItemsWarningMessage": "PICK UP ALL GROUND ITEMS! They will be cleared in {0} seconds.",
  "RestartingNowMessage": "Server restarting now.",
  "ExitCode": 0
}
```

The actual process restart depends on the external process manager (systemd, Docker, etc.). The server calls `Stop()` with a configurable exit code.

### Fixed on review

- The pre-clear warning only got scheduled when `ClearGroundItemsLeadSeconds` fit before the restart's total seconds. A lead time configured longer than the restart itself (e.g. lead 120s on `/restart 1`) silently skipped the warning while still clearing on schedule, destroying items with zero notice. Fixed by clamping the delay to 0 instead of skipping the callback, so the warning always fires, at worst simultaneously with the clear.
- `ExecuteStop`'s cleanup call had no try/catch, unlike the sibling `StratumItemCleanup.DoCleanup`. An exception during `Die()` would propagate out of the `RegisterCallback`-invoked `ExecuteStop` before reaching `server.Stop()`, so a scheduled restart could silently never happen. Now caught and logged, restart still proceeds.
- `/clearitems` had no way to cancel its 30-second warning once issued, the one asymmetry against `/restart cancel`. Added `/clearitems cancel`.
- Collapsed three separate `ColorizeVtml` implementations into the one `StratumChatFormatter` already exposes, and pulled the ground-item sweep into a shared, age-parameterized static method on `StratumItemCleanup` so the restart path stops duplicating the `LoadedEntities` filter and `Die()` loop (deliberately passes minimum age 0, since `/restart`/`/clearitems` clear everything after their own warning, unlike the periodic pass which only takes abandoned items).

### Testing detail

Both build passes clean (`dotnet build VintageStory.slnx -c Release`, `-p:EmbedPatchedFiles=true`). Smoke-tested with startup commands: `/restart 1` shows countdown, `/clearitems now` runs cleanup, `/restart cancel` aborts the scheduled restart. Not tested with a connected client end to end.

## Type

- [ ] Bug fix
- [ ] Performance
- [x] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

## Related issues

Fixes #219

### DIFF
--- a/sources/VintagestoryLib/Vintagestory.Server/CmdStratumStaffCommands.cs
+++ b/sources/VintagestoryLib/Vintagestory.Server/CmdStratumStaffCommands.cs
@@ -290,6 +290,24 @@ internal class CmdStratumStaffCommands
 				.HandleWith(HandleReports);
 		}
 
+		if (StratumCommandRegistration.ShouldRegister(commands.Restart, "/restart", "Commands.Restart"))
+		{
+			server.api.commandapi.Create("restart")
+				.WithDescription("Schedule a server restart with countdown warnings")
+				.WithArgs(parsers.Word("minutes or cancel"))
+				.RequiresPrivilege(Privilege.chat)
+				.HandleWith(HandleRestart);
+		}
+
+		if (StratumCommandRegistration.ShouldRegister(commands.ClearItems, "/clearitems", "Commands.ClearItems"))
+		{
+			server.api.commandapi.Create("clearitems")
+				.WithDescription("Clear all dropped items on the ground")
+				.WithArgs(parsers.OptionalWordRange("mode", "now"))
+				.RequiresPrivilege(Privilege.chat)
+				.HandleWith(HandleClearItems);
+		}
+
 		StratumTargetCommandOverrides.Register(server);
 	}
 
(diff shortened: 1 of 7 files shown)


Zaldaryon:
Found and fixed four issues on review, e24fb5c:

1. **Real bug**: the pre-clear warning only got scheduled when `ClearGroundItemsLeadSeconds` fit before the restart's total seconds. A lead time configured longer than the restart itself (e.g. lead 120s on `/restart 1`) silently skipped the warning while still clearing on schedule, destroying items with zero notice. Fixed by clamping the delay to 0 instead of skipping the callback, so the warning always fires, at worst simultaneously with the clear.
2. **Real bug**: `ExecuteStop`'s cleanup call had no try/catch, unlike the sibling `StratumItemCleanup.DoCleanup`. An exception during `Die()` would propagate out of the `RegisterCallback`-invoked `ExecuteStop` before reaching `server.Stop()`, so a scheduled restart could silently never happen. Now caught and logged, restart still proceeds.
3. `/clearitems` had no way to cancel its 30-second warning once issued, the one asymmetry against `/restart cancel`. Added `/clearitems cancel`.
4. Collapsed three separate `ColorizeVtml` implementations into the one `StratumChatFormatter` already exposes, and pulled the ground-item sweep into a shared, age-parameterized static method on `StratumItemCleanup` so the restart path stops duplicating the `LoadedEntities` filter and `Die()` loop (deliberately passes minimum age 0, since `/restart`/`/clearitems` clear everything after their own warning, unlike the periodic pass which only takes abandoned items).

Both builds (`dotnet build -c Release` and `-p:EmbedPatchedFiles=true`) clean on the updated branch.

PR #241, opened 2026-08-09
Add /wilderness command

### LINKED ISSUE #217: /wilderness command
### Problem

On overpopulated large survival servers, players do not have an easy way to get away from spawn after their introductions to the server. Filled with tons of claimed chunks and cities around spawn that they can’t/don’t partake in, that could extend out for multiple thousands of blocks.

### Proposed change

/wilderness command that teleports the player to a random part in the world that is not a claimed area. Can be configured to be a random location within X radius of spawn, so random within 1000 blocks of spawn, 5000, 10,000 etc.

### Alternatives considered

_No response_

### Scope

Zaldaryon:
## Summary

Adds `/wilderness`, a player command that teleports to a random unclaimed spot within a radius band of spawn.

`Commands.WildernessSettings` (default `MinRadiusBlocks` 1000, `MaxRadiusBlocks` 5000, `MaxAttempts` 40) is a min/max band rather than a single radius, so a server can push players out past its claimed belt instead of teleporting into the middle of it. Both bounds and the attempt budget are configurable per the issue's "1000/5000/10000 etc." example.

Candidate search rides `ServerMain.LocateRandomPosition`, the same chunk-queueing search vanilla uses for random player-spawn placement, instead of a new hand-rolled chunk-loading path. Candidates far from spawn are normally in unloaded chunks; `LocateRandomPosition` already knows how to queue a load and re-test once it lands. A candidate is accepted only if:

- `ServerSystemSupplyChunks.AdjustForSaveSpawnSpot` (vanilla's own surface-Y + non-liquid + solid-ground helper, also used for spawn placement) resolves it to safe ground, called with `forPlayer: null` so a claim that happens to grant this specific player access still counts as claimed
- it's outside `MinRadiusBlocks` of spawn, checked on the position `AdjustForSaveSpawnSpot` resolved to (the one actually used as the teleport target, not the pre-adjustment candidate)
- `World.Claims.Get` at the resolved position returns no claims (the authoritative "is this actually claimed" check, done separately from `AdjustForSaveSpawnSpot`'s own internal claim check since that one is access-based rather than ownership-based)

Because the search is asynchronous relative to the command call, `HandleWilderness` returns immediately with a "searching" message; `StratumWildernessSystem` tracks one in-flight search per player UID so spamming `/wilderness` can't pile up searches. Exhausting `MaxAttempts` reports failure to the player rather than looping forever or teleporting into a claim.

The actual teleport goes through `StratumTeleportWarmups.StartOrTeleport`, exactly like `/home` and `/spawn`, so the existing warmup/cancel-on-move/cancel-on-damage UX is inherited rather than reimplemented.

Gated on a new `stratum.wilderness` privilege via the standard `StratumCommandAccessConfig` (auto-granted to normal survival roles by default, same as `/home`/`/spawn`), catalog entry added to `StratumCommandAccessCatalog`. Cooldown is configurable but 0 by default, matching the rest of that command family.

### Fixed on review

`AdjustForSaveSpawnSpot` mutates the candidate's X/Z in place while walking to solid ground or off liquid, up to a small random walk per retry, most likely to trigger right at the edge of the claimed spawn sprawl `MinRadiusBlocks` exists to route players past (that's exactly where the first surface probe tends to hit liquid or a claim). The `MinRadiusBlocks` check originally ran before that adjustment, on the pre-adjustment position, so a candidate accepted past the minimum radius could walk back inside it during adjustment and still get used as the teleport target, silently defeating the one guarantee `/wilderness` exists to provide. Fixed by reordering: adjust first, then check the radius on the position that's actually going to be used.

One non-blocking item not acted on: `FindOnlineClient` re-implements a lookup pattern that already exists in `CmdStratumEssentials.cs` and `StratumTeleportWarmups.cs`, a fourth copy of pre-existing duplication. Left as is, extracting all four call sites is out of scope for this PR.

### Testing detail

Both build passes clean, 0 errors, 0 new warnings:
```
dotnet build VintageStory.slnx -c Release
dotnet build VintageStory.slnx -c Release -p:EmbedPatchedFiles=true
```
`scripts/smoke-test.sh` passes (server reaches `RunGame`).

Not tested with an actual connected client running `/wilderness`. No client/account was available in this session, so the in-game search/warmup/teleport/claim-rejection behavior is verified by code review and the boot smoke test only, not an end-to-end run.

## Type

- [ ] Bug fix
- [ ] Performance
- [x] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

## Related issues

Fixes #217

### DIFF
--- a/sources/VintagestoryLib/Vintagestory.Server/CmdStratumEssentials.cs
+++ b/sources/VintagestoryLib/Vintagestory.Server/CmdStratumEssentials.cs
@@ -52,6 +52,14 @@ internal class CmdStratumEssentials
 				.HandleWith(HandleSetSpawn);
 		}
 
+		if (StratumCommandRegistration.ShouldRegister(commands.Wilderness, "/wilderness", "Commands.Wilderness"))
+		{
+			server.api.commandapi.Create("wilderness")
+				.WithDescription("Teleport to a random unclaimed spot near spawn")
+				.RequiresPrivilege(Privilege.chat)
+				.HandleWith(HandleWilderness);
+		}
+
 		if (StratumCommandRegistration.ShouldRegister(commands.TeleportRequests.Request, "/tpa commands", "Commands.TeleportRequests.Request"))
 		{
 			server.api.commandapi.Create("tpa")
@@ -202,6 +210,22 @@ internal class CmdStratumEssentials
 
 		StratumRuntime.LogInfo($"spawn set by {player.PlayerName} at {pos.X}, {pos.Y}, {pos.Z}");
 		return TextCommandResult.Success(StratumCommandText.Confirm("Spawn set", "to " + pos.X + ", " + pos.Y + ", " + pos.Z + "."));
+	}
+
+	private TextCommandResult HandleWilderness(TextCommandCallingArgs args)
+	{
+		if (!CheckAccess(args, StratumRuntime.Config.Commands.Wilderness, "wilderness", out TextCommandResult failure))
+		{
+			return failure;
+		}
+
+		IServerPlayer player = GetPlayer(args);
+		if (player == null)
+		{
+			return TextCommandResult.Error("Only players can use /wilderness.");
+		}
+
+		return StratumWildernessSystem.Request(server, player);
 	}
 
 	private TextCommandResult HandleTpa(TextCommandCallingArgs args)
--- a/sources/VintagestoryLib/Vintagestory.Server/StratumCommandAccessCatalog.cs
+++ b/sources/VintagestoryLib/Vintagestory.Server/StratumCommandAccessCatalog.cs
@@ -32,6 +32,7 @@ internal static class StratumCommandAccessCatalog
 	{
 		yield return new StratumCommandAccessEntry("spawn", "/spawn", commands.Spawn, "Use /spawn");
 		yield return new StratumCommandAccessEntry("setspawn", "/setspawn", commands.SetSpawn, "Use /setspawn");
+		yield return new StratumCommandAccessEntry("wilderness", "/wilderness", commands.Wilderness, "Use /wilderness");
 		yield return new StratumCommandAccessEntry("tpa", "/tpa", commands.TeleportRequests.Request, "Use /tpa, /tpaccept, /tpdeny, and /tpacancel");
 		yield return new StratumCommandAccessEntry("tpahere", "/tpahere", commands.TeleportRequests.Here, "Use /tpahere");
 		yield return new StratumCommandAccessEntry("home", "/home", commands.Homes.Home, "Use /home and /homes");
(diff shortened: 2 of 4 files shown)


Zaldaryon:
Found and fixed a real bug on review, 0909073:

`AdjustForSaveSpawnSpot` mutates the candidate's X/Z in place while walking to solid ground or off liquid, up to a small random walk per retry, most likely to trigger right at the edge of the claimed spawn sprawl `MinRadiusBlocks` exists to route players past (that's exactly where the first surface probe tends to hit liquid or a claim). The `MinRadiusBlocks` check ran before that adjustment, on the pre-adjustment position, so a candidate accepted past the minimum radius could walk back inside it during adjustment and still get used as the teleport target, silently defeating the one guarantee `/wilderness` exists to provide.

Fixed by reordering: adjust first, then check the radius on the position that's actually going to be used.

One non-blocking finding not acted on: `FindOnlineClient` re-implements a lookup pattern that already exists in `CmdStratumEssentials.cs` and `StratumTeleportWarmups.cs`, a fourth copy of pre-existing duplication. Left as is, extracting all four call sites is out of scope for this PR.

Both builds (`dotnet build -c Release` and `-p:EmbedPatchedFiles=true`) verified clean on the updated branch.

Zaldaryon:
Re-audited with live Atlas validation and found a second, more serious bug, fixed in `6b305fa`.

`AdjustForSaveSpawnSpot` was called with `forPlayer: null`, a deliberate choice so a claim that only grants this player access still counts as "claimed". But `WorldMap.GetBlockingLandClaimant(forPlayer, pos, accessFlag)` has a fallback specific to the null-player case: once a map region has ever held a claim, its `LandClaimByRegion` entry is emptied on removal but the dictionary key is never deleted, so every position in that region gets treated as blocked by a sentinel `ServerLandClaim` forever after — not just the area the original claim covered. On any server that's ever had a land claim anywhere near a `/wilderness` search's candidate area (realistically most servers that have actually been played on), this can silently exhaust the whole `MaxAttempts` budget and report failure even where plenty of genuinely open, unclaimed land exists.

Found it by adding a second Atlas fixture that claims the entire search area and expects rejection (`WildernessScenarios.Wilderness_request_never_teleports_into_a_fully_claimed_search_area`): after clearing every claim on the disposable test world, the search still failed every single time, tracing back to this fallback rather than any real claim. Ruled out claims (`World.Claims.All` confirmed empty), terrain (`AdjustForSaveSpawnSpot`/`GetTerrainGenSurfacePosY` both succeed on the same coordinates called directly), and the teleport-target lookup (`FindOnlineClient` resolves fine) one at a time before landing on this.

Fixed by passing the real player instead of `null`, matching vanilla's own `SpawnPlayerRandomlyAround` (which always passes the joining player, never null). This doesn't weaken the "avoid claimed land even with access" guarantee the null choice was meant to provide — the explicit ownership-based `World.Claims.Get` check a few lines below already enforces that independently.

Also fixed a false positive in the existing minimum-radius Atlas scenario while investigating: it measured the player's post-command distance from `World.Spawn`, but the test world's `spawnRadius` join jitter can already place a freshly joined player far enough from spawn to satisfy that check before `/wilderness` ever runs, so it had never actually proven a teleport happened. Rewritten to measure movement from the player's own post-join position instead.

Both Atlas scenarios (minimum-radius search, claim-area rejection) now pass genuinely against a live patched server.

Issue #242, opened 2026-08-09
StratumCombatLogSystem constructed three times in ServerSystemCommands.cs


Zaldaryon:
### Problem

\`patches/VintagestoryLib/Vintagestory.Server/ServerSystemCommands.cs.patch\` (from the recent PvP combat log prevention change) adds three identical lines:

\`\`\`csharp
new StratumCombatLogSystem(server);
new StratumCombatLogSystem(server);
new StratumCombatLogSystem(server);
\`\`\`

instead of one. Confirmed in the actual generated build tree after a fresh bootstrap, not just the patch file.

### Effect

Three separate \`StratumCombatLogSystem\` instances each register their own \`OnPlayerInteractEntity\`, \`OnPlayerDisconnect\`, \`OnPlayerJoin\` handlers and a 1-second game tick listener, and each construction overwrites the static \`StratumCombatLogHook.OnProjectilePvPHit\` delegate (only the last instance's handler stays wired for the projectile path). Every melee/projectile PvP hit runs the tag logic three times: harmless for the tag dictionary itself (just re-set to the same value), but \`BroadcastMessageToAllGroups\`, staff alerts, and the tag/expiry chat messages all fire three times per event once a player is actually tagged. The kill-on-disconnect penalty also runs through three separate \`OnPlayerDisconnect\` handlers, though \`Entity.Die\` looks idempotent against \`Alive\`, so it's more visible noise than a correctness break there.

### Where

\`patches/VintagestoryLib/Vintagestory.Server/ServerSystemCommands.cs.patch\`, three consecutive \`new StratumCombatLogSystem(server);\` lines. Fix is deleting two of them.

Found opportunistically while implementing #215 in the same construction block, filed separately since it's unrelated.

Zaldaryon:
The fix is in #249. It keeps one StratumCombatLogSystem instance in ServerSystemCommands, so combat interaction, join, and disconnect handlers register once. The PR includes the Atlas regression and native smoke results. Fixes #242 in the PR will close this issue when the PR merges.

Zaldaryon:
The fix from #249 is now in indev at b0414a0. ServerSystemCommands constructs one StratumCombatLogSystem, so combat interaction, join, and disconnect handlers register once. This resolves #242. I am closing the issue because the merged change removes the duplicate registrations.

PR #243, opened 2026-08-09
Add /pvecombat cooperative combat toggle

### LINKED ISSUE #215: Cooperative PVE combat toggle
### Problem

During PVE combat players are not able to attack the same creature as each other at the same time. This proves harder for stronger mobs while there are multiple players able to attack it at once, but they can only fight one at a time.

### Proposed change

Lower the immunity frames for creatures through the toggle pve cooperative combat command. Which enables players to cooperate on pve combat against a single creature at one time 

### Alternatives considered

_No response_

### Scope

Zaldaryon:
## Summary

Adds `/pvecombat on|off|toggle|status`, a config-backed toggle that lowers (or removes) the post-hit invulnerability window on creatures so a group can cooperatively damage the same target instead of one attacker's hit blocking everyone else's.

`EntityBehaviorHealth` (VSEssentials) puts every entity that takes damage on a hardcoded 500ms `invulnerable` activity timer. That's fine for a single attacker, but it means a group fighting one strong creature only ever lands one effective hit every 500ms combined, not per player: whichever attacker's hit lands first inside that window is the only one that counts, everyone else's simultaneous or near-simultaneous hits do zero damage.

`StratumCoopCombatHook` is a small static holder added to `VintagestoryAPI`, the one assembly both `VintagestoryLib` (where Stratum's server config lives) and `VSEssentials` (where `EntityBehaviorHealth` lives) can reach, since neither references the other directly. `EntityBehaviorHealth` reads `CreatureInvulnerableMs` from it instead of the literal 500 whenever the entity taking damage isn't a player, so this never touches PvP or a player's own hit-stun, only creature-side invulnerability. `StratumCoopCombatSystem` owns the value, applies it from config at boot, and the command flips it live, persisted through the existing `SaveConfig` path the same way `/chattoggle` is.

Config lives in a new `CoopCombat` section:
```json
{
  "Enabled": false,
  "CreatureInvulnerableMs": 100
}
```
`CreatureInvulnerableMs` is clamped 0-500. Off by default, vanilla 500ms behavior untouched until a server opts in.

### Anticheat interaction checked

`StratumCombatGuard`'s kill-aura heuristic keys on how many distinct entities *one attacker* hits inside a short window (`MultiTargetThreshold`), not on how many attackers hit *one entity*. Cooperative PvE combat is several players sharing one target, so it never touches that heuristic, each attacker still only ever hits the one creature.

### Also found, unrelated

While adding this system's construction line next to the existing combat log system's in `ServerSystemCommands.cs.patch`, found `StratumCombatLogSystem` is constructed three times there instead of once (pre-existing, from the PvP combat log prevention change already on `indev`). Left untouched and filed separately: #242.

### Testing detail

Both build passes clean (`dotnet build VintageStory.slnx -c Release`, `-p:EmbedPatchedFiles=true`), 0 errors. `scripts/smoke-test.sh` passes (server reaches `RunGame`). `/pvecombat` itself not tested with a connected client, no account available this session; command registration and config wiring verified by code review and the boot smoke test only.

## Type

- [ ] Bug fix
- [ ] Performance
- [x] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

## Related issues

Fixes #215

### DIFF
--- a/patches/VSEssentials/Entity/Behavior/BehaviorHealth.cs.patch
+++ b/patches/VSEssentials/Entity/Behavior/BehaviorHealth.cs.patch
@@ -1,0 +1,34 @@
+diff --git a/VSEssentials/Entity/Behavior/BehaviorHealth.cs b/VSEssentials/Entity/Behavior/BehaviorHealth.cs
+index c961b3a..ee4c4cb 100644
+--- a/VSEssentials/Entity/Behavior/BehaviorHealth.cs
++++ b/VSEssentials/Entity/Behavior/BehaviorHealth.cs
+@@ -8,10 +8,11 @@ using Vintagestory.API.Client;
+ using Vintagestory.API.Common;
+ using Vintagestory.API.Common.Entities;
+ using Vintagestory.API.Config;
+ using Vintagestory.API.Datastructures;
+ using Vintagestory.API.MathTools;
++using Vintagestory.API.Server; // Stratum: cooperative PvE combat toggle hook
+ using Vintagestory.API.Util;
+ 
+ namespace Vintagestory.GameContent;
+ 
+ public interface ICanHealCreature
+@@ -235,11 +236,16 @@ public class EntityBehaviorHealth : EntityBehavior
+                 entity.Attributes.SetInt("dmgkb", 1);
+             }
+ 
+             if (damage > 0.05)
+             {
+-                entity.SetActivityRunning("invulnerable", 500);
++                // Stratum: cooperative PvE combat toggle shortens (or removes) this window for
++                // non-player entities only, so several players can land hits on the same
++                // creature inside the window that used to block every attacker but the first.
++                // Players keep the vanilla 500ms always, this never touches PvP or self-hits.
++                int invulnerableMs = entity is EntityPlayer ? 500 : StratumCoopCombatHook.CreatureInvulnerableMs;
++                entity.SetActivityRunning("invulnerable", invulnerableMs);
+                 // Gets already called on the server directly
+                 if (entity.World.Side == EnumAppSide.Client)
+                 {
+                     entity.OnHurt(null, entity.WatchedAttributes.GetFloat("onHurt", 0));
+                 }
--- a/patches/VintagestoryLib/Vintagestory.Server/ServerSystemCommands.cs.patch
+++ b/patches/VintagestoryLib/Vintagestory.Server/ServerSystemCommands.cs.patch
@@ -2,7 +2,7 @@ diff --git a/VintagestoryLib/Vintagestory.Server/ServerSystemCommands.cs b/Vinta
 index 1d4be19..e8394a8 100644
 --- a/VintagestoryLib/Vintagestory.Server/ServerSystemCommands.cs
 +++ b/VintagestoryLib/Vintagestory.Server/ServerSystemCommands.cs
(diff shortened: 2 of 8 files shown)


Zaldaryon:
Found and fixed two real bugs on review, 7e66090:

1. \`HandleToggle\` only gated on \`Privilege.chat\`, the base privilege nearly every player has, and never checked the \`stratum.coopcombat\` privilege the command's own config wires up. Any regular player could flip a server-wide combat-balance setting and it would persist via \`SaveConfig\`. Fixed by adding the same \`CheckAccess\` pattern (\`Commands.Enabled\`, \`access.Enabled\`, \`CallerHasAccess\`, cooldown) every sibling command handler already uses.
2. \`Apply()\` was documented as re-running on a live \`/stratum reload\` but was never actually wired into \`CmdStratum.HandleReload\`, so \`StratumCoopCombatHook.CreatureInvulnerableMs\` stayed stale after a config-file edit and reload, even though \`/pvecombat status\` would report the new value correctly since it reads live config, not the hook. Fixed by calling \`StratumCoopCombatSystem.Apply(StratumRuntime.Config.CoopCombat)\` in the reload success path, alongside the other reload-time re-sync steps.

Both builds (\`dotnet build -c Release\` and \`-p:EmbedPatchedFiles=true\`) clean on the updated branch.

Zaldaryon:
Re-audited with live Atlas validation (previously untested end to end) plus an independent Opus review pass, and found one more real bug before merging.

`StratumCommandsConfig.EnsurePopulated()` wires up every sibling command's access config with both a `??=` null-coalescing default and an `.EnsurePopulated("privilege.string")` call — `CoopCombat` had neither. `StratumCommandAccessCatalog.CallerHasAccess` treats a null/blank `Privilege` as "everyone has access", so a `stratum-commands.json` with a `CoopCombat` entry present but missing `Privilege` (hand-edited, or written by an older build) would silently reopen the exact permission bypass this PR already found and fixed once in `HandleToggle`. Fixed by adding the same two lines every other command gets.

Added two Atlas scenarios to close the coverage gap on both previously-self-reviewed fixes, since neither had a regression test:
- `Pvecombat_command_is_blocked_when_its_access_config_is_disabled` — also asserts `Privilege == "stratum.coopcombat"` directly, guarding the bug above.
- `Stratum_reload_resyncs_the_damage_hook_from_persisted_config` — drifts the hook from persisted config directly and requires `/stratum reload` to resync it, rather than trusting the doc comment.

All three Atlas scenarios pass against a live patched server.

Two non-blocking notes from the review, not acted on:
- **Client/server hook divergence.** Stock clients never write `CreatureInvulnerableMs`, so a client always holds 500ms locally while the server applies the configured value. `EntityAgent`'s hit-sound/particle gating reads the same `IsActivityRunning("invulnerable")` flag, so on a dedicated server, players in a group get no local feedback for 2nd+ hits the server actually counted. Cosmetic, not a correctness bug, but worth knowing if reports come in about "hits not registering" that the server logs as landing.
- **Scope slightly broader than described.** The hook affects every non-player entity taking damage from any source not setting `IgnoreInvFrames` (`AiTaskMeleeAttack`/`AiTaskShootAtEntity` already default it true), which includes player melee and player projectile damage, not just the multi-attacker case in the PR description — weapon durability drains proportionally faster per real hit landed while enabled, since durability loss is gated on the same `!IsActivityRunning` check. Likely intended, just undocumented.

Merged into `indev`.

PR #260, opened 2026-08-16
feat: support linux-arm64

HanosThePlacid:
## Summary

Adds Linux ARM64 support to the Stratum launcher while keeping the existing x64 and Windows paths unchanged.

## Changes

- Publishes a Linux ARM64 apphost from the release workflow.
- Downloads the official Linux x64 server archive for the managed server files and portable assets.
- Pins the ARM64 native overlay by Vintage Story major and minor version in `forks.json`.
- Embeds the pinned catalog in the launcher, so startup does not depend on the GitHub API or an unauthenticated fallback.
- Verifies the overlay archive with its pinned SHA256 digest before extraction.
- Copies only native shared libraries from the overlay. Managed files and the overlay host executable are not copied because they can lag the selected base patch.
- Applies the native overlay even when the base install marker already exists, including an install moved from x64 to ARM64.
- Records the base version, asset name, and digest in `.stratum-arm64-natives`.
- Removes the unused x86-64 `libopenal.so.1` from ARM64 installs.
- Verifies that all required native libraries exist and that every top-level native library in `Lib` is an AArch64 ELF.
- Rebases the branch onto the current `indev` branch.

## Checklist

- [x] Linux ARM64 apphost cross-publish
- [x] Pinned overlay URLs and SHA256 digests
- [x] Missing overlay mapping fails closed
- [x] Missing or malformed digest fails closed
- [x] Existing-install and marker refresh paths handled
- [x] No managed files copied from the overlay
- [x] Documentation updated

## Verification

- `bash scripts/bootstrap.sh`
- `dotnet build VintageStory.slnx -c Release`
- `dotnet build VintageStory.slnx -c Release -p:EmbedPatchedFiles=true`
- `bash scripts/smoke-test.sh`, reached `RunGame` without fatal errors
- `scripts/pack-release.ps1` cross-publish for `linux-arm64`
- Published launcher inspected as an ARM AArch64 ELF
- Pinned 1.21 and 1.22 overlay archives matched their SHA256 values and contained the expected AArch64 libraries
- Official 1.22.7 Linux x64 archive inspected to confirm the x86-64 natives replaced by the launcher

## Related issues

Closes #258

### DIFF
--- a/StratumServer/VanillaBootstrap.cs
+++ b/StratumServer/VanillaBootstrap.cs
@@ -2,7 +2,6 @@ using System;
 using System.IO;
 using System.IO.Compression;
 using System.Net.Http;
-using System.Net.Http.Headers;
 using System.Runtime.InteropServices;
 using System.Security.Cryptography;
 using System.Text.Json;
@@ -16,7 +15,9 @@ internal static class VanillaBootstrap
 	private const string VersionManifestUrl = "https://api.vintagestory.at/stable-unstable.json";
 
 	// Anego publishes the aarch64 natives as a standalone overlay; the main manifest is x64-only.
-	private const string Arm64ReleasesUrl = "https://api.github.com/repos/anegostudios/VintagestoryServerArm64/releases?per_page=50";
+	// Pinned in forks.json (arm64NativeOverlays), embedded into the assembly - resolving it needs
+	// no GitHub API call at runtime.
+	private const string Arm64CatalogResourceName = "Stratum.forks.json";
 
 	internal static void EnsureVanillaAssets(string baseGameVersion, bool refresh)
 	{
@@ -98,7 +99,7 @@ internal static class VanillaBootstrap
 		// Swap the handful of architecture-specific natives before PatchedFileOverlay runs.
 		if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux) && RuntimeInformation.ProcessArchitecture == Architecture.Arm64)
 		{
-			ApplyArm64NativeOverlay(installDir, cacheDir, baseGameVersion, overwriteExisting);
+			ApplyArm64NativeOverlay(installDir, cacheDir, baseGameVersion, refresh);
 		}
 
 		File.WriteAllText(markerPath, expectedMarker);
@@ -153,12 +154,12 @@ internal static class VanillaBootstrap
 	// patch release or two, so copying its managed/host files would risk a version mismatch against
 	// Stratum's patched assemblies. The natives are self-contained third-party libraries
 	// (SkiaSharp, e_sqlite3, zstd) whose ABI does not track Vintage Story patch releases.
-	private static void ApplyArm64NativeOverlay(string installDir, string cacheDir, string baseGameVersion, bool overwriteExisting)
+	private static void ApplyArm64NativeOverlay(string installDir, string cacheDir, string baseGameVersion, bool refresh)
 	{
 		string markerPath = Path.Combine(installDir, ".stratum-arm64-natives");
-		if (!overwriteExisting && File.Exists(markerPath))
+		if (!refresh && File.Exists(markerPath) && File.ReadAllText(markerPath).Trim() == baseGameVersion)
 		{
-			Console.WriteLine("Stratum: arm64 natives already in place; leaving them untouched");
+			Console.WriteLine("Stratum: arm64 natives already in place for " + baseGameVersion + "; leaving them untouched");
 			return;
 		}
 
@@ -224,7 +225,7 @@ internal static class VanillaBootstrap
 		}
 
 		Console.WriteLine($"Stratum: replaced {copied} native librar(ies) with aarch64 builds from {asset.Name}");
-		File.WriteAllText(markerPath, asset.Name);
+		File.WriteAllText(markerPath, baseGameVersion);
 	}
 
 	private static bool IsNativeLibrary(string path)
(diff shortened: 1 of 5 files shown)


StratumServer/VanillaBootstrap.cs
```diff
@@ -136,6 +147,199 @@ private static ArchiveInfo GetArchiveForPlatform(string version)
...
+	// Replaces the x86-64 natives from the official archive with the aarch64 builds Anego publishes
+	// as a separate overlay release. Only *.so files are taken: everything else in that overlay is
+	// either portable IL or VintagestoryServer host files, and the overlay lags the base game by a
+	// patch release or two, so copying its managed/host files would risk a version mismatch against
+	// Stratum's patched assemblies. The natives are self-contained third-party libraries
+	// (SkiaSharp, e_sqlite3, zstd) whose ABI does not track Vintage Story patch releases.
+	private static void ApplyArm64NativeOverlay(string installDir, string cacheDir, string baseGameVersion, bool overwriteExisting)
+	{
+		string markerPath = Path.Combine(installDir, ".stratum-arm64-natives");
+		if (!overwriteExisting && File.Exists(markerPath))
```
trevorftp: This never fires? `overwriteExisting` is always true by the time this method runs, seems like it only gets called when `EnsureVanillaAssets` didn't do its early return at line 30, and that early return only skips when refresh is true or `markerMatches` is false. Either of which forces` overwriteExisting = true` at line 36.

So this check can't ever take the skip branch? also every refresh would redo the GitHub API and re-verifies even if we had the right overlay already installed. I would probably want to check the marker before deciding to touch the network at all, which is the same as the .stratum-base check up top

StratumServer/VanillaBootstrap.cs
```diff
@@ -136,6 +147,199 @@ private static ArchiveInfo GetArchiveForPlatform(string version)
...
+
+	private static string MajorMinor(string version)
+	{
+		string[] parts = version.Split('.');
+		return parts.Length >= 2 ? parts[0] + "." + parts[1] : version;
+	}
+
+	private static bool VerifyArm64Digest(string path, string digest)
+	{
+		if (string.IsNullOrWhiteSpace(digest))
```
trevorftp: This seems like it would just skip verification instead of failing when GitHub's asset JSON has no digest field, which probably isn't guaranteed on every release. Your main archive path a few methods up treats a bad or missing checksum as fatal. And that means this path is doing the same job, downloading code from a third-party repo that gets loaded into the server process, so I'd want it at least as strict.

StratumServer/VanillaBootstrap.cs
```diff
@@ -14,6 +15,9 @@ internal static class VanillaBootstrap
 {
 	private const string VersionManifestUrl = "https://api.vintagestory.at/stable-unstable.json";
 
+	// Anego publishes the aarch64 natives as a standalone overlay; the main manifest is x64-only.
+	private const string Arm64ReleasesUrl = "https://api.github.com/repos/anegostudios/VintagestoryServerArm64/releases?per_page=50";
```
trevorftp: Would rather see this pinned like the rest of our deps in `forks.json` than resolved dynamically

StratumServer/VanillaBootstrap.cs
```diff
@@ -136,6 +147,199 @@ private static ArchiveInfo GetArchiveForPlatform(string version)
...
+			}
+		}
+
+		if (preferred.Name != null)
+		{
+			return preferred;
+		}
+		if (fallback.Name != null)
+		{
+			Console.WriteLine($"Stratum: no arm64 overlay published for {MajorMinor(baseGameVersion)}.x; falling back to {fallback.Name}");
```
trevorftp: Worth a second look regardless of my pin question above, this falls back to installing whatever overlay is newest even if it doesn't match the base game's minor version

trevorftp (dismissed):

pizza2004:
> Anego publishes no linux-arm64 server archive

Do you think we should be?

HanosThePlacid:
Pushed a fix for all three review points (commit 8bc451f (https://github.com/HanosThePlacid/StratumArm64/commit/8bc451f)):

1. Unreachable skip check — overwriteExisting was always true on every path that reached ApplyArm64NativeOverlay (EnsureVanillaAssets already returns earlier for the "nothing changed" case), so !overwriteExisting could never be true. Replaced with a marker check against baseGameVersion, gated on refresh, so a version bump or explicit --stratum-refresh still reapplies the overlay, but nothing else does.

2. Weak digest verification — VerifyArm64Digest now throws on a missing or malformed digest instead of logging a warning and treating the file as verified.

3. Dynamic GitHub API resolution — Dropped the runtime GitHub API call entirely. anegostudios/VintagestoryServerArm64 release assets (name/url/sha256) are now pinned in forks.json under arm64NativeOverlays, embedded into the StratumServer assembly at build time, and read back via GetManifestResourceStream — no network round-trip (or GITHUB_TOKEN/rate-limit concern) before the digest check even runs.

Verified on real aarch64 hardware via Pterodactyl (Debian 13, .NET 10), tag v1.22.6-stratum.3-arm64test.1

Zaldaryon (approved):
Reviewed the rebased head d9b867686a27bcee69aa7779ae3c132e95deb957. The requested bootstrap changes are addressed: arm64 overlay resolution is pinned by minor version, digest verification fails closed, existing x64 installs receive the overlay on arm64, the marker records the exact asset and digest, and remaining native libraries are validated as AArch64. The Linux ARM64 package cross-publish and the embedded-file smoke test passed, with the server reaching RunGame and no fatal errors. No automated CI status is published for this PR.

PR #261, opened 2026-08-16
Microblock Optimization and GenRockStrataNew

tehtelev:
### BlockEntityMicroBlock

**ConvertToVoxels:**
- ThreadLocal caches for `BoolArray16x16x16` and `byte[,,]` — eliminated per-call object allocations.
- `Span<byte>` replaces 3D indexing of `byte[,,]` (removed multiplications + 3 bounds checks per element).
- `SetRangeTrue` / `Fill` instead of nested Z-loop assignments.

**RebuildCuboidList:**
- `VoxelVisited` from ThreadLocal — zero allocations during rebuild.
- `Span<byte> flatMat` + `GetFlat`/`SetFlatTrue` replaces 3D access to `VoxelMaterial`.
- Fixed buffer `uint[4096]` with `cuboidCount++` instead of `List.Add`.
- Precomputed flags (`isWest`, `isEast`, `xCenter`, `yCenter`) hoisted out of inner loops.

**TryGrowX / TryGrowY / TryGrowZ:**
- Switched to `Span<byte>` and flat indexing.
- Base offsets (`baseX`/`baseXY`/`baseY`) computed once before nested loops.

> **Summary:** eliminated GC allocations during voxel rebuild, improved memory locality via flat access and `Span<T>`.

---

### BoolArray16x16x16 (new class)
- Stores 4096 bits in an array of 64 `ulong`s (512 bytes).
- `GetFlat`, `SetFlatTrue`, `SetRangeTrue` — fast bitwise ops without branch checks.
- `[x,y,z]` indexer preserved for backward compatibility.

> **Summary:** compact storage + O(1) bitwise operations instead of bool array.

---

### BEBehaviorMicroblockSnowCover

**buildSnowCuboids:**
- `Span<bool>` (stackalloc 256) replaces heap-allocated `bool[16,16]` — removed GC pressure per call.
- Flat indexing for visited and voxel access (`dx * 256 + yOffset + dz`).
- Removed unused iteration over `aboveCuboids.Contains()`.

**TrySnowableSurfaceGrowX / TrySnowableSurfaceGrowZ:**
- `Span<bool>` replaces `bool[,]` for visited.
- Flat indexing for voxels (`idx + 16` for Y+1).

> **Summary:** eliminated heap allocations in the hot snow-cover path

---

### `GenRockStrataNew optimization`

Implemented batch reading and writing of data in chunks. Why? This reduces the number of write locks and reduces the overhead of block-by-block access to palette data. Corresponding methods have also been added to `ChunkData` and `ChunkDataLayer` to implement this functionality. They may also be useful elsewhere.
Added corresponding overloads for the `IChunkBlocks` interface.

## Type

- [ ] Bug fix
- [x] Performance
- [ ] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

### `Microblock optimization`

Tested on the generation of 31417 chunk columns (6 threads):
- BlockEntityMicroBlock.FromTreeAttributes method execution time: reduced from 9904 ms to 5228 ms;
- BlockEntityMicroBlock GC memory allocations: reduced from 2.47 Gb to 1.57 GB.

**Before (vanilla)**
microblock before trace
microblock before mem

**After**
microblock after trace
microblock after mem

### `GenRockStrataNew optimization`

Tested on the generation of 31417 chunk columns (6 threads):
- GenRockStrataNew.StratumGenChunkColumn method execution time: reduced from 113547 ms to 96604 ms;
- GenRockStrataNew.genBlockColumn method execution time: reduced from 113426 ms to 57765 ms;

**Before**
image

**After**
image

### DIFF
--- a/patches/VSSurvivalMod/Systems/Microblock/BEBehaviorMicroblockSnowCover.cs.patch
+++ b/patches/VSSurvivalMod/Systems/Microblock/BEBehaviorMicroblockSnowCover.cs.patch
@@ -1,0 +1,207 @@
+diff --git a/VSSurvivalMod/Systems/Microblock/BEBehaviorMicroblockSnowCover.cs b/VSSurvivalMod/Systems/Microblock/BEBehaviorMicroblockSnowCover.cs
+index 61d9053..a16a99f 100644
+--- a/VSSurvivalMod/Systems/Microblock/BEBehaviorMicroblockSnowCover.cs
++++ b/VSSurvivalMod/Systems/Microblock/BEBehaviorMicroblockSnowCover.cs
+@@ -2,10 +2,12 @@ using System;
+ using System.Collections.Generic;
+ using Vintagestory.API.Client;
+ using Vintagestory.API.Common;
+ using Vintagestory.API.Datastructures;
+ using Vintagestory.API.MathTools;
++using System.Runtime.CompilerServices;
++using System.Runtime.InteropServices;
+ 
+ #nullable disable
+ 
+ namespace Vintagestory.GameContent
+ {
+@@ -65,52 +67,51 @@ namespace Vintagestory.GameContent
+         }
+ 
+ 
+         CuboidWithMaterial[] aboveCuboids = null;
+ 
++        [MethodImpl(MethodImplOptions.AggressiveInlining)] // Stratum: added to hint compiler to inline the hot method
+         void buildSnowCuboids(BoolArray16x16x16 Voxels)
+         {
+             List<uint> snowCuboids = new List<uint>();
+             List<uint> groundSnowCuboids = new List<uint>();
+ 
+             var aboveBe = Api?.World.BlockAccessor.GetBlockEntity(Pos.UpCopy()) as BlockEntityMicroBlock;
+             CuboidWithMaterial[] newAboveCuboids = null;
+ 
+-           
+-            bool[,] snowVoxelVisited = new bool[16, 16];
++            // Stratum start: Replaced bool[16,16] heap allocation with stackalloc Span<bool> to avoid per-call GC pressure.
++            // Also flattened the 2D voxel/visited access into single-dimension indices for better cache locality.
++            Span<bool> snowVoxelVisited = stackalloc bool[256];
+ 
+             for (int dy = 15; dy >= 0; dy--)
+             {
++                int yOffset = dy * 16; // flat index base for the entire Y plane within Voxels
++
+                 for (int dx = 0; dx < 16; dx++)
+                 {
++                    int flatBase = dx * 256 + yOffset;     // base for Voxels.GetFlat(flatBase + dz)
++                    int visitedRowBase = dx * 16;           // base for snowVoxelVisited[dx*16+dz]
++
+                     for (int dz = 0; dz < 16; dz++)
+                     {
+-                        if (snowVoxelVisited[dx, dz]) continue;
+-                        bool ground = dy == 0 && !Voxels[dx, dy, dz];
+-                        bool search = ground || Voxels[dx, dy, dz];
++                        int visitedIdx = visitedRowBase + dz;
++                        if (snowVoxelVisited[visitedIdx]) continue;
++
++                        bool voxelHere = Voxels.GetFlat(flatBase + dz);
++                        bool ground = dy == 0 && !voxelHere;
++                        bool search = ground || voxelHere;
+ 
+                         if (!search) continue;
+ 
+-                        if (dy == 15)
++                        if (dy == 15 && aboveBe != null && newAboveCuboids == null)
+                         {
+-                            if (aboveBe != null)
++                            newAboveCuboids = new CuboidWithMaterial[aboveBe.VoxelCuboids.Count];
++                            for (int i = 0; i < newAboveCuboids.Length; i++)
+                             {
+-                                if (newAboveCuboids == null)
+-                                {
+-                                    newAboveCuboids = new CuboidWithMaterial[aboveBe.VoxelCuboids.Count];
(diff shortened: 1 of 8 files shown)


patches/VSSurvivalMod/Systems/Microblock/BEMicroBlock.cs.patch
```diff
@@ -1,8 +1,394 @@
...
++        // ThreadLocal for BoolArray, byte[,,] and VoxelVisited — avoid new on every RebuildCuboidList; fixed uint[] buffer 4096 for cuboids (max 16³)
++        protected static readonly ThreadLocal<BoolArray16x16x16> RebuildVoxelsTL =
++            new ThreadLocal<BoolArray16x16x16>(() => new BoolArray16x16x16());
++        protected static readonly ThreadLocal<byte[,,]> RebuildMaterialsTL =
++            new ThreadLocal<byte[,,]>(() => new byte[16, 16, 16]);
++        protected static readonly ThreadLocal<BoolArray16x16x16> VoxelVisitedTL =
++            new ThreadLocal<BoolArray16x16x16>(() => new BoolArray16x16x16());
++
++        // Fixed buffer for cuboids (max 16*16*16 = 4096)
++        private static readonly uint[] CuboidBuffer = new uint[4096];
```
trevorftp: This one's a plain static array, not `ThreadLocal` like` RebuildVoxelsTL`/`RebuildMaterialsTL`/`VoxelVisitedTL` right above it. If `RebuildCuboidList` can run on two threads at once, which seems plausible given worldgen here is already multi-threaded, both would write into the same array using independent `cuboidCount` counters starting at 0, and potentially cause issues with each other? Should this be `ThreadLocal<uint[]>` too?

patches/VSSurvivalMod/Systems/Microblock/BEMicroBlock.cs.patch
```diff
@@ -1,8 +1,394 @@
...
+ 
+         public void ConvertToVoxels(out BoolArray16x16x16 voxels, out byte[,,] materials)
+         {
+-            voxels = new BoolArray16x16x16();
+-            materials = new byte[16, 16, 16];
++            // Stratum start:
++            // instead of new BoolArray/byte[][] we use ThreadLocal + Span<byte> for flat addressing of byte[,,]; SetRangeTrue/Fill instead of nested loops
++            voxels = RebuildVoxelsTL.Value;
++            voxels.Clear();
++            materials = RebuildMaterialsTL.Value;
```
trevorftp, line 42: voxels gets `.Clear()`'d right above but materials doesn't get reset before reuse. I would think cells that end up outside any cuboid keep whatever a previous block entity on that thread last wrote there. I mean I think this is harmless now, but surely could be an issue for anything added later that reads materials unconditionally. Might be worth a comment noting this, or just clearing it!

patches/VSSurvivalMod/Systems/Microblock/BEMicroBlock.cs.patch
```diff
@@ -1,8 +1,394 @@
...
++                            didGrowAny |= TryGrowZ(cub, Voxels, VoxelVisited, flatMat);
+                         }
+ 
+-                        voxelCuboids.Add(ToUint(cub));
++                        CuboidBuffer[cuboidCount++] = ToUint(cub);
+                     }
+                 }
+             }
+ 
+-            this.VoxelCuboids = voxelCuboids;
```
trevorftp, line 241: This line's deleted and nothing replaces it? 

`RebuildCuboidList` builds the new shape into `CuboidBuffer`/`cuboidCount` now, but `this.VoxelCuboids` never gets updated from it, unless im missing something. Every call here is void and relied on this method updating the instances as a side effect, which is same as before. I think this means the visual/collision shape goes stale after any edit that calls `RebuildCuboidList`,  (e.g. chiseling, snow cover growth, anything through the overload). Needs a `VoxelCuboids = new List<uint>(CuboidBuffer.AsSpan(0, cuboidCount).ToArray())` or similar at the end.

patches/VSSurvivalMod/Systems/Microblock/BEBehaviorMicroblockSnowCover.cs.patch
```diff
@@ -0,0 +1,207 @@
...
+ 
+ namespace Vintagestory.GameContent
+ {
+@@ -65,52 +67,51 @@ namespace Vintagestory.GameContent
+         }
+ 
+ 
+         CuboidWithMaterial[] aboveCuboids = null;
+ 
++        [MethodImpl(MethodImplOptions.AggressiveInlining)] // Stratum: added to hint compiler to inline the hot method
```
trevorftp, line 24: This is really minor but, this method has nested loops and multiple early-outs, the JIT's inlining heuristics are very unlikely to honor this hint regardless of the attribute. I really don't think this is doing anything, but whatever tyou want to do is up to you.

trevorftp (changes requested):
Overall solid optimization, 3 of the 4 comments I would say are potentially issues. 1 of them is a nitpick, keep up the solid work 👍

tehtelev:
1. Thanks for the tip. This is a real race condition with multithreaded worldgen. Fixed.
2. Thanks for pointing this out. Currently, it's only safe because consumers only read material at positions with voxel == true. But for safety and clarity, I added a cleanup.
3. Thanks, you're absolutely right—this looks like a critical bug. RebuildCuboidList writes to CuboidBuffer, but VoxelCuboids isn't updated. Fixed.
4. I confirm—VoxelCuboids should be synchronized with the buffer at the end of RebuildCuboidList. Fixed.
5. I didn't touch it. It won't affect performance in any way.

PR #266, opened 2026-08-19
Read and write batching for lighting calculations and more

tehtelev:
## Summary

Continued work on chunk data batching.
- The light system has been switched to reading and writing light in batches.
- Branches using AVX2 processor instructions have been added to the chunkdatalayer for bulk data reading and writing. These greatly speed up data handling. If the processor doesn't support them, the code will execute as usual without them.
- The chunk data writing system has been improved in GenRockStrataNew. Now, all changes are written directly to the snapshot and then bulk written to the chunk.
- Fixed a bug where ProPickWorkSpace was causing memory leaks because it didn't take into account the batching system in GenRockStrataNew.

Write batching for GenRockStrata has been improved.

## Type

- [ ] Bug fix
- [x] Performance
- [ ] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

## Performance numbers

Tested on the generation of 31417 chunk columns (6 threads):
- GenRockStrataNew.StratumGenChunkColumn method execution time: reduced from 88880 ms to 65377 ms;
- SunFloodChunkColumnForWorldGen method execution time: reduced from 117987 ms to 87647 ms;

**Before**
image
image

**After**
image
image

### DIFF
--- a/patches/VSEssentials/Systems/WorldGen/Standard/ChunkGen/2.GenRockStrata/GenRockStrata.cs.patch
+++ b/patches/VSEssentials/Systems/WorldGen/Standard/ChunkGen/2.GenRockStrata/GenRockStrata.cs.patch
@@ -1,5 +1,5 @@
 diff --git a/VSEssentials/Systems/WorldGen/Standard/ChunkGen/2.GenRockStrata/GenRockStrata.cs b/VSEssentials/Systems/WorldGen/Standard/ChunkGen/2.GenRockStrata/GenRockStrata.cs
-index cb9eefa..3b30334 100644
+index cb9eefa..574c98e 100644
 --- a/VSEssentials/Systems/WorldGen/Standard/ChunkGen/2.GenRockStrata/GenRockStrata.cs
 +++ b/VSEssentials/Systems/WorldGen/Standard/ChunkGen/2.GenRockStrata/GenRockStrata.cs
 @@ -1,6 +1,7 @@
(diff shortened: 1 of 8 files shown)


Zaldaryon:
@tehtelev Ran the correctness half of this independently, plus a hardware question the description doesn't answer: what happens on a machine without AVX2. Short version: the batching is correct everywhere I could test it, and it surfaces a real crash in the new bulk write path that's worth fixing before merge.

## AVX2 vs. scalar fallback: byte-identical

Built `ChunkDataLayer.cs` from this branch standalone and ran a 30-fixture round-trip battery (`SetBlocksBulkUnsafe` → `CopyBlocksToUnsafe`/`CopyLightToUnsafe`) twice: once natively, once with `DOTNET_EnableAVX2=0` forcing every `Avx2.IsSupported` branch in `DecodePlanesAvx2` and `IsUniformWord` to its scalar fallback. Fixtures covered all-air, uniform, low and high cardinality (up to 10,000 distinct block values in one subchunk), three palette-growth-timing variants, and 20 randomized realistic-terrain fixtures. Diffing the two runs, hashes included, is empty. `CopyLightToUnsafe` also matches `CopyBlocksToUnsafe` exactly on every fixture, confirming the shared decode path works the same for light as for blocks.

## A crash, not a truncation

`SetBlocksBulkUnsafe`'s new `PaletteIndexForBulk` grows the palette with no equivalent to the guard vanilla's own `MakeSpaceInPalette` → `CleanUpPalette` already has. On a client-pooled chunk (`pool.server == null`), vanilla refuses to grow the palette past 14 bit-planes and throws a deliberate, player-facing exception instead: *"Oops, a client chunk had so many changes that it exceeded the maximum size..."* `PaletteIndexForBulk` has no such check. Pushed to true capacity (32,768 distinct block values in one subchunk, one past the point vanilla's own guard already stops at), it throws a raw, unhandled `IndexOutOfRangeException` from `ChunkDataLayer.cs:1981` (`dataBits[bitsize] = pool.NewData();`), indexing past the 15-slot `dataBits` array. Confirmed identical with AVX2 on and off, so it's a write-path gap, not a SIMD issue.

Whether a single 32×32×32 subchunk realistically accumulates that many distinct values in play is genuinely unclear. Natural rock strata generation stayed at 3 to 16 distinct types across the randomized fixtures I ran, but `ProPickWorkSpace` and long-lived, heavily modded or heavily built client chunks seem like a more plausible route there over time, which is exactly the scenario vanilla's own guard exists for.

## Performance holds, but is far more AVX2-dependent than the numbers above show

Old per-position `Get()` loop vs. `CopyBlocksToUnsafe`, same fixture, JIT-warmed:

| | AVX2 native | `DOTNET_EnableAVX2=0` |
| --- | --- | --- |
| Stopwatch, 20k decodes | 4.50x | 1.34x |
| dotTrace Sampling, 150k decodes | 5.55x | 1.10x |

decode-speedup


Servers on non-AVX2 hardware, or on the `linux-arm64` build #260 is adding (where `Avx2.IsSupported` is always `false`, being an X86-only intrinsics class), should expect something closer to 1.1x to 1.3x, not the 26 to 35 percent the screenshots above show. Worth a line in the PR description so nobody's surprised later.

Not run: real linux-arm64 hardware, and the full in-game world-gen determinism regression. Both need the actual game runtime and, for arm64, hardware this pass didn't have access to.

tehtelev:
Added checks to avoid IndexOutOfRangeException. Everything is fine.

PR #268, opened 2026-08-19
Add automatic punishment ladder for repeat anticheat violations

### LINKED ISSUE #221: Cheating Player Punishments
### Problem

Be able to config more things to inhibit cheating players

### Proposed change

When a cheating player is flagged enough times and determined to be cheating. Can have a config that can wipe their inventory (or drop their inventory), freeze them and teleport them to jail before they are banned from the server. 

### Alternatives considered

_No response_

### Scope

Zaldaryon:
## Summary

Adds an opt-in escalation ladder for players who keep accumulating confirmed anticheat flags: inventory drop/wipe, then freeze, then jail, then a timed ban, in that order, based on a persistent per-player flag count that survives relogs and restarts.

Every action the ladder takes reuses an existing, already-tested primitive:
- Inventory drop/wipe: `InventoryBase.DropAll`/`Clear`, same as normal death handling, skipping the creative inventory the same way `EntityPlayer.WalkInventory` already does.
- Freeze: `StratumStaffCommandState.Freeze`, the same as `/freeze`.
- Jail: a new `CmdStratumStaffCommands.JailAutomatically` entry point that reuses the existing `TryGetJailLocation` + `JailSingleTarget` methods unchanged, reached through a minimal `internal static Instance` reference since the active-jail tracking and enforcement tick loop are instance state with no prior external access point.
- Ban: `PlayerDataManager.BanPlayer` + `DisconnectPlayer`, same as `/ban`.

Nothing new is invented for enforcement; only the threshold ladder and the persistent standing behind it are new.

Standing is tracked separately from `StratumAnticheatReporter`'s existing in-memory `PlayerViolations` dictionary, which is intentionally rolling and prunes a player out after `KeepPlayerViolationsMinutes` of inactivity. A new `StratumAnticheatHistory` store persists `TotalFlags` and the highest tier already applied in `CustomPlayerData`, the same pattern `StratumModerationStore`/`StratumCustodyStore` already use, so a repeat offender who logs off can't reset their standing by relogging.

The escalation check runs synchronously right after `RecordViolation`'s lock releases, in the same place every `Record*Violation` call site already calls `DisconnectPlayer` directly for `KickConfirmedCheats` today. That existing pattern already proves the calling context is safe for real game-state actions, not only logging, so this needed no new queue or deferred-execution scheduling.

`/stratum ac <player>` now also shows accumulated flags and the next tier due. `StratumRuntime`'s config safety check gained four warnings for risky combinations: punishments enabled while anticheat itself is disabled, wipe-instead-of-drop, a very low first threshold, and a permanent (non-recovering) ban duration.

Off by default (`Anticheat.Punishments.Enabled = false`), same posture as `KickConfirmedCheats` per rule: an operator has to opt in with eyes open, since chaining these into automatic consequences on a false positive is a materially different risk than an alert or a single kick.

## Type

- [x] New feature

## Checklist

- [x] `scripts/extract-patches.sh` ran clean (no vanilla-file changes; every edit here is under `sources/`, no `patches/` diffs touched).
- [x] `dotnet build VintageStory.slnx -c Release` is green (0 errors, 0 new warnings; verified no warnings in any touched file).
- [x] Every vanilla edit has a `// Stratum` marker. (N/A: no vanilla files touched, only Stratum-original files under `sources/`.)
- [x] Tested on a real server start, not just compilation (`make smoke`: PASS, reached `WorldReady`, no fatal errors).

## Related issues

Fixes #221

### DIFF
--- a/sources/VintagestoryLib/Vintagestory.Server/CmdStratum.cs
+++ b/sources/VintagestoryLib/Vintagestory.Server/CmdStratum.cs
@@ -243,7 +243,7 @@ internal class CmdStratum
 
 	private TextCommandResult HandleAnticheat(string playerName)
 	{
-		return TextCommandResult.Success(StratumAnticheatReporter.BuildReport(playerName));
+		return TextCommandResult.Success(StratumAnticheatReporter.BuildReport(server, playerName));
 	}
 
 	private TextCommandResult HandleHealth()
--- a/sources/VintagestoryLib/Vintagestory.Server/CmdStratumStaffCommands.cs
+++ b/sources/VintagestoryLib/Vintagestory.Server/CmdStratumStaffCommands.cs
@@ -17,6 +17,13 @@ internal class CmdStratumStaffCommands
 {
 	private const string VanishIndicatorCode = "stratum-vanish-active";
 
+	// Constructed exactly once per server (ServerSystemCommands.cs.patch), never held onto by its
+	// caller. StratumAnticheatPunishments needs the jail mechanism this class owns (the active-jail
+	// tracking set and the tick loop that enforces it are both instance state, not reusable statics
+	// the way Freeze and BanPlayer already are), so this is the smallest way to reach it from
+	// outside without duplicating that tracking elsewhere.
+	internal static CmdStratumStaffCommands Instance { get; private set; }
+
 	private readonly ServerMain server;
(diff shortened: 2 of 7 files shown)


tehtelev (commented):
Looks good. No complaints.

Issue #203, opened 2026-07-28
Game Teams


PHoenixOPHury:
### Problem

Groups feature in game is very limited in what you can do with it and the players within that group. It is limited to people joining it, having a group chat, sharing map markers and syncing land claims I believe. 

### Proposed change

Teams would be a separate feature from groups. Teams would allow you to add players to it through commands or a button on the UI. Moderators on the server can modify the teams (or whatever role has permission set to be able to modify them in the config). Be able to turn off friendly fire to where players on the same team can no longer damage each other. This enables team based things to be done in game without the issue of accidentally killing your teammate which happens a lot during PVP or even PVE when attacking the same creature. Be able to see which players are on each team. Be able to set kits for the team, as in a gear kit like putting specific tools, weapons, items or armor that the players on that team either spawn with or can do the kit command to activate that teams kit. Teams should have a team chat. Maybe a thing that displays on their names what team they are on, could be color based (like red circle etc). Be able to set a spawn point for the team. Make announcements to that specific team, an announcement that appears both in chat and over the players screen as text. 

### Scope

- [x] This is a server-side change. It does not require a custom client.

trevorftp:
This just does not feel in the scope of what Stratum is or does. Belongs more in a server/client mod.

PR #269, opened 2026-08-19
Add live kit creator: /kit and /kitedit

### LINKED ISSUE #210: Kit Creator
### Problem

Kits cannot be made without a mod, it’s complicated to make, takes a while, no UI, having to know the exact name of items to put into it otherwise the kit won’t work. Having to do it through json and cannot do it while the server is running. 

### Proposed change

Be able to edit the kit in game through commands and a UI possibly, while the server is running. Assign the kits to roles. Much more easily put and take out items from the kit. Be able to activate the kit through a command or through respawn, which is configurable. The kits should be able to include weapons, tools, armor, food, backpacks and other items. 

### Alternatives considered

_No response_

### Scope

Zaldaryon:
## Summary

Kits editable in-game, live, via commands, with role assignment and configurable activation, closing #210. A UI isn't possible here (the issue's own scope checkbox says server-side only, no custom client), so this is a command surface, same as the rest of Stratum.

The issue's real complaint was having to know exact item codes and edit JSON by hand while the server is offline. Both `/kitedit create <name>` and `/kitedit additem <name>` solve that by snapshotting a live `ItemStack` straight off the caller (their whole inventory, or their active hotbar slot) instead of asking for a code anywhere.

**Giving items** (`StratumKitGiver`) reuses existing primitives rather than inventing new ones:
- Armor goes into the character inventory explicitly, matched by `ItemSlotCharacter`'s own dress-type compatibility check (`CanHold`); this doesn't duplicate that check, it just calls it.
- Everything else goes through `Entity.TryGiveItemStack`, the same entry point the existing `/giveitem` command already uses in production.
- A worn backpack's own contents already live inside its `ItemStack`'s attribute tree (`CollectibleBehaviorHeldBag` stores them under a `"backpack"` tree on the bag's own stack), so giving a filled backpack round-trips its contents in one piece. No separate "give backpack, then fill it" step needed.
- If nothing fits anywhere, the item drops at the player's feet rather than vanishing.

**Storage** (`StratumKitStore`): kits are global, staff-authored operational data, not tunable settings, so they get their own `stratum-kits.json` sidecar, loaded once at boot from `CmdStratumKits`'s constructor, rather than folding into `StratumRuntime`'s config load/save cycle the way `Commands`/`Performance` do.

**Commands** (`CmdStratumKits`): `/kit [name]` redeems a kit you're assigned (or lists what you can use); `/kitedit` covers `create`, `additem`, `removeitem`, `delete`, `rename`, `setrole` (toggles a role scope on/off), `setcooldown`, `onrespawn`, `oneperlife`, `preview`, `give`, `list`. `AssignedTo` is a general scope list (`"role:x"`) rather than a role-only field, so #203's later team kits shouldn't need a rewrite here.

**Give-on-respawn** polls for the player actually being revived before giving, rather than giving synchronously inside the respawn event handler. Vanilla's own handler for that event only starts an async teleport, calling `Entity.Revive()` later once the target chunk finishes loading, so giving immediately would race that revive and could lose the items.

## A real bug found writing test coverage

While writing Atlas scenarios for the drop-at-feet fallback, found that `PlayerInventoryManager.TryGiveItemstack` zeroes out whatever stack it was handed once its own internal transfer slot ends up empty, even when the overall give reports failure (a partial absorption elsewhere that doesn't count as a completed transfer). The original code reused that same stack reference for the drop-at-feet fallback, which would have silently discarded the item under exactly the circumstance that fallback exists to prevent. Fixed by giving a clone and dropping whatever it has left afterward.

## Type

- [x] New feature

## Checklist

- [x] `scripts/extract-patches.sh` ran clean. One intentional line in `ServerSystemCommands.cs.patch` registering the new command class, nothing else touched.
- [x] `dotnet build VintageStory.slnx -c Release` is green (0 errors, 0 new warnings; verified no warnings in any touched file).
- [x] Every vanilla edit has a `// Stratum` marker. The one vanilla-file line (`ServerSystemCommands.cs`) carries `// Stratum: issue #210 kit creator`.
- [x] Tested on a real server start, not just compilation (`make smoke`: PASS, reached `WorldReady`).

## Testing

Beyond build and smoke test, this got real functional coverage: 4 new Atlas scenarios (headless server-tick-driven tests, not just compile checks) covering armor placement into the character inventory, a filled backpack's contents surviving the give, the drop-at-feet fallback when every inventory is full, JSON persistence (upsert/rename/delete), the full `/kitedit` command surface end-to-end, and give-on-respawn with role-scope gating and the one-per-life marker. All 15 scenarios in the existing Atlas suite pass, no regressions. This is private tooling, not part of Stratum's own repo or CI, so it isn't part of this PR's diff, but it's what actually caught the `TryGiveItemstack` bug above.

## Related issues

Fixes #210

### DIFF
--- a/sources/VintagestoryLib/Vintagestory.Server/CmdStratumKits.cs
+++ b/sources/VintagestoryLib/Vintagestory.Server/CmdStratumKits.cs
@@ -2,8 +2,11 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 using Vintagestory.API.Common;
 using Vintagestory.API.Config;
+using Vintagestory.API.Datastructures;
 using Vintagestory.API.Server;
 
 namespace Vintagestory.Server;
@@ -29,10 +32,7 @@ internal class CmdStratumKits
 		StratumKitStore.Load();
 		server.EventManager.OnPlayerRespawn += OnPlayerRespawn;
 
-		if (!StratumRuntime.Config.Commands.Enabled)
-		{
-			return;
-		}
+		if (!StratumRuntime.Config.Commands.Enabled) return;
 
 		CommandArgumentParsers parsers = server.api.commandapi.Parsers;
 
@@ -54,9 +54,7 @@ internal class CmdStratumKits
 				.WithArgs(
 					parsers.WordRange("action", "create", "additem", "removeitem", "delete", "rename", "setrole", "setcooldown", "onrespawn", "oneperlife", "preview", "give", "list"),
 					parsers.OptionalWord("name"),
-					parsers.OptionalWord("value"),
-					parsers.OptionalWord("value2"))
-				.RequiresPrivilege(Privilege.gamemode)
+					parsers.OptionalWord("value"))
 				.HandleWith(HandleKitEdit);
 		}
 	}
@@ -66,17 +64,14 @@ internal class CmdStratumKits
 	// itself is keyed to.
 	private void OnPlayerRespawn(IServerPlayer player)
 	{
-		if (player?.ServerData == null)
+		if (player?.ServerData == null) return;
+
+		if (player.ServerData.CustomPlayerData.Remove(RedeemedThisLifeKey))
 		{
-			return;
+			server.PlayerDataManager.playerDataDirty = true;
 		}
 
-		player.ServerData.CustomPlayerData.Remove(RedeemedThisLifeKey);
-
-		if (!StratumRuntime.Config.Commands.Kits.Enabled)
-		{
-			return;
-		}
+		if (!StratumRuntime.Config.Commands.Kits.Enabled) return;
 
 		// This event fires before the player is actually revived: vanilla's own handler for it
 		// (ServerSystemEntitySimulation.OnPlayerRespawn) only starts an async teleport, calling
@@ -91,16 +86,10 @@ internal class CmdStratumKits
 		{
 			ConnectedClient client = server.GetClientByUID(playerUid);
 			bool alive = client?.Player?.Entity?.Alive == true;
-			if (!alive && client != null && --attemptsRemaining > 0)
-			{
-				return;
-			}
+			if (!alive && client != null && --attemptsRemaining > 0) return;
 
 			server.UnregisterGameTickListener(listenerId);
-			if (client?.Player == null || !alive)
-			{
-				return;
-			}
+			if (client?.Player == null || !alive) return;
 
 			GiveAssignedRespawnKits(client.Player, client.ServerData);
 		}, 250);
@@ -110,7 +99,9 @@ internal class CmdStratumKits
 	{
 		foreach (StratumKitDefinition kit in StratumKitStore.All())
 		{
-			if (kit.GiveOnRespawn && IsAssignedTo(kit, data))
+			if (kit.GiveOnRespawn
+				&& IsAssignedTo(kit, data)
+				&& StratumCommandAccessCatalog.PlayerHasAccess(target, StratumRuntime.Config.Commands.Kits))
 			{
 				StratumKitGiver.Give(target, kit);
 				MarkRedeemedThisLife(data, kit.Name);
@@ -203,7 +194,6 @@ internal class CmdStratumKits
 		string action = (args[0] as string)?.ToLowerInvariant();
 		string name = args.Parsers[1].IsMissing ? null : args[1] as string;
 		string value = args.Parsers[2].IsMissing ? null : args[2] as string;
-		string value2 = args.Parsers[3].IsMissing ? null : args[3] as string;
 
 		return action switch
 		{
@@ -481,8 +471,9 @@ internal class CmdStratumKits
 		List<StratumKitItem> items = new List<StratumKitItem>();
 		foreach (InventoryBase inventory in player.InventoryManager.InventoriesOrdered)
 		{
+			// A worn backpack's contents are stored in the backpack stack in the character inventory.
+			// Walking the backpack inventory too would add every content item a second time.
 			if (inventory.ClassName != GlobalConstants.hotBarInvClassName
-				&& inventory.ClassName != GlobalConstants.backpackInvClassName
 				&& inventory.ClassName != GlobalConstants.characterInvClassName)
 			{
 				continue;
@@ -502,9 +493,27 @@ internal class CmdStratumKits
 
 	private static StratumKitItem EncodeStack(ItemStack stack)
 	{
+		JsonItemStack jsonStack = new JsonItemStack
+		{
+			Type = stack.Class,
+			Code = stack.Collectible?.Code,
+			StackSize = stack.StackSize,
+			Attributes = stack.Attributes == null ? null : JsonObject.FromJson(stack.Attributes.ToJsonToken())
+		};
+		JObject json = new JObject
+		{
+			["type"] = jsonStack.Type == EnumItemClass.Block ? "block" : "item",
+			["code"] = jsonStack.Code?.ToShortString(),
+			["stacksize"] = jsonStack.StackSize
+		};
+		if (jsonStack.Attributes?.Token != null)
+		{
+			json["attributes"] = jsonStack.Attributes.Token.DeepClone();
+		}
+
 		return new StratumKitItem
 		{
-			Base64 = Convert.ToBase64String(stack.Clone().ToBytes()),
+			StackJson = json.ToString(Formatting.None),
 			Code = stack.Collectible?.Code?.ToString() ?? "unknown",
 			Quantity = stack.StackSize
 		};
@@ -544,6 +553,12 @@ internal class CmdStratumKits
 	private bool CheckAccess(TextCommandCallingArgs args, string commandLabel, StratumCommandAccessConfig access, out TextCommandResult failure)
 	{
 		failure = null;
+		if (!StratumRuntime.Config.Commands.Enabled)
+		{
+			failure = TextCommandResult.Error("Stratum commands are disabled.");
+			return false;
+		}
+
 		if (access == null || !access.Enabled)
 		{
 			failure = TextCommandResult.Error("/" + commandLabel + " is disabled.");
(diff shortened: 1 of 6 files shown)


sources/VintagestoryLib/Vintagestory.Server/CmdStratumKits.cs
```diff
@@ -0,0 +1,567 @@
...
+	private readonly ServerMain server;
+
+	public CmdStratumKits(ServerMain server)
+	{
+		this.server = server;
+		StratumRuntime.Config.EnsurePopulated();
+		StratumKitStore.Load();
+		server.EventManager.OnPlayerRespawn += OnPlayerRespawn;
+
+		if (!StratumRuntime.Config.Commands.Enabled)
```
trevorftp: `if (!StratumRuntime.Config.Commands.Enabled) return;` would look cleaner.

sources/VintagestoryLib/Vintagestory.Server/CmdStratumKits.cs
```diff
@@ -0,0 +1,567 @@
...
+				.HandleWith(HandleKitEdit);
+		}
+	}
+
+	// #210: "activate the kit ... through respawn, which is configurable". Also clears the
+	// one-per-life redeemed set for every kit, the same life boundary the give-on-respawn pass
+	// itself is keyed to.
+	private void OnPlayerRespawn(IServerPlayer player)
+	{
+		if (player?.ServerData == null)
```
trevorftp: Same as above, 1 line pls

sources/VintagestoryLib/Vintagestory.Server/CmdStratumKits.cs
```diff
@@ -0,0 +1,567 @@
...
+		{
+			return;
+		}
+
+		player.ServerData.CustomPlayerData.Remove(RedeemedThisLifeKey);
+
+		if (!StratumRuntime.Config.Commands.Kits.Enabled)
```
trevorftp: Here too

sources/VintagestoryLib/Vintagestory.Server/CmdStratumKits.cs
```diff
@@ -0,0 +1,567 @@
...
+		// for Alive first instead, giving up after ~10s if it never arrives (e.g. the player
+		// disconnected mid-respawn).
+		string playerUid = player.PlayerUID;
+		int attemptsRemaining = 40;
+		long listenerId = 0;
+		listenerId = server.RegisterGameTickListener(_ =>
+		{
+			ConnectedClient client = server.GetClientByUID(playerUid);
+			bool alive = client?.Player?.Entity?.Alive == true;
+			if (!alive && client != null && --attemptsRemaining > 0)
```
trevorftp: Here and one below too

sources/VintagestoryLib/Vintagestory.Server/CmdStratumKits.cs
```diff
@@ -0,0 +1,567 @@
...
+			: TextCommandResult.Success(StratumCommandText.Warning("Kit '" + kit.Name + "' partially given: " + summary));
+	}
+
+	private static List<StratumKitItem> SnapshotInventory(IPlayer player)
+	{
+		List<StratumKitItem> items = new List<StratumKitItem>();
+		foreach (InventoryBase inventory in player.InventoryManager.InventoriesOrdered)
+		{
+			if (inventory.ClassName != GlobalConstants.hotBarInvClassName
+				&& inventory.ClassName != GlobalConstants.backpackInvClassName
```
trevorftp: Would this cause duplication? This inventory includes the worn bag slots and their contents, but the bag stack already stores those contents.

sources/VintagestoryLib/Vintagestory.Server/StratumKitGiver.cs
```diff
@@ -0,0 +1,148 @@
...
+	{
+		if (string.IsNullOrWhiteSpace(entry?.Base64))
+		{
+			return null;
+		}
+
+		try
+		{
+			byte[] data = Convert.FromBase64String(entry.Base64);
+			ItemStack stack = new ItemStack(data);
```
trevorftp, line 92: Could this use JsonItemStack? Raw ItemStack bytes store the numeric registry ID, so kits can resolve to the wrong item if the registry IDs change

sources/VintagestoryLib/Vintagestory.Server/StratumConfig.cs
```diff
@@ -1635,6 +1635,10 @@ internal class StratumCommandsConfig
 
 	public StratumCommandAccessConfig ClearItems { get; set; } = StratumCommandAccessConfig.ForPrivilege("stratum.clearitems");
 
+	public StratumCommandAccessConfig Kits { get; set; } = StratumCommandAccessConfig.ForPrivilege("stratum.kits");
```
trevorftp, line 1638: Kits and KitsEdit need entries in StratumCommandAccessCatalog too, their privileges don't go through normal registration.

sources/VintagestoryLib/Vintagestory.Server/CmdStratumKits.cs
```diff
@@ -0,0 +1,567 @@
...
+
+			GiveAssignedRespawnKits(client.Player, client.ServerData);
+		}, 250);
+	}
+
+	private void GiveAssignedRespawnKits(ServerPlayer target, ServerPlayerData data)
+	{
+		foreach (StratumKitDefinition kit in StratumKitStore.All())
+		{
+			if (kit.GiveOnRespawn && IsAssignedTo(kit, data))
```
trevorftp: This skips the stratum.kits access check. With no role assigned, a respawn kit goes to everyone instead of everyone with the kit privilege

sources/VintagestoryLib/Vintagestory.Server/CmdStratumKits.cs
```diff
@@ -0,0 +1,567 @@
...
+		playerData.CustomPlayerData.TryGetValue(RedeemedThisLifeKey, out string raw);
+		HashSet<string> redeemed = string.IsNullOrWhiteSpace(raw)
+			? new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+			: new HashSet<string>(raw.Split(';'), StringComparer.OrdinalIgnoreCase);
+		redeemed.Add(kitName);
+		playerData.CustomPlayerData[RedeemedThisLifeKey] = string.Join(";", redeemed);
+		server.PlayerDataManager.playerDataDirty = true;
+	}
+
+	private bool CheckAccess(TextCommandCallingArgs args, string commandLabel, StratumCommandAccessConfig access, out TextCommandResult failure)
```
trevorftp, line 553: This is missing Commands.Enabled check the other command handlers have.

sources/VintagestoryLib/Vintagestory.Server/CmdStratumKits.cs
```diff
@@ -0,0 +1,567 @@
...
+		if (StratumCommandRegistration.ShouldRegister(kitEditAccess, "/kitedit", "Commands.KitEdit"))
+		{
+			server.api.commandapi.Create("kitedit")
+				.WithDescription("Create and manage kits: /kitedit create|additem|removeitem|delete|rename|setrole|setcooldown|onrespawn|oneperlife|preview|give|list <...>")
+				.WithArgs(
+					parsers.WordRange("action", "create", "additem", "removeitem", "delete", "rename", "setrole", "setcooldown", "onrespawn", "oneperlife", "preview", "give", "list"),
+					parsers.OptionalWord("name"),
+					parsers.OptionalWord("value"),
+					parsers.OptionalWord("value2"))
+				.RequiresPrivilege(Privilege.gamemode)
```
trevorftp: Why require gamemode here? stratum.kitedit already handles access, and this stops roles with that custom privilege from using the command.

sources/VintagestoryLib/Vintagestory.Server/CmdStratumKits.cs
```diff
@@ -0,0 +1,567 @@
...
+	{
+		if (!CheckAccess(args, "kitedit", StratumRuntime.Config.Commands.KitEdit, out TextCommandResult failure))
+		{
+			return failure;
+		}
+
+		string action = (args[0] as string)?.ToLowerInvariant();
+		string name = args.Parsers[1].IsMissing ? null : args[1] as string;
+		string value = args.Parsers[2].IsMissing ? null : args[2] as string;
+		string value2 = args.Parsers[3].IsMissing ? null : args[3] as string;
```
trevorftp: value2 doesnt seem to be used anymore?

sources/VintagestoryLib/Vintagestory.Server/CmdStratumKits.cs
```diff
@@ -0,0 +1,567 @@
...
+	// one-per-life redeemed set for every kit, the same life boundary the give-on-respawn pass
+	// itself is keyed to.
+	private void OnPlayerRespawn(IServerPlayer player)
+	{
+		if (player?.ServerData == null)
+		{
+			return;
+		}
+
+		player.ServerData.CustomPlayerData.Remove(RedeemedThisLifeKey);
```
trevorftp: Should mark playerDataDirty when this actually removes the key or I would this this may not persist

trevorftp (changes requested):

Zaldaryon:
Addressed the requested review changes in ae865be.

New kit entries store JsonItemStack codes and attributes. The decoder keeps the Base64 path for kits created by earlier builds. Inventory snapshots skip the separate backpack inventory because the worn backpack stack stores its contents.

The access catalog includes /kit and /kitedit. Respawn delivery checks stratum.kits access. The global Commands.Enabled guard applies to every kit command. /kitedit no longer requires gamemode privilege or parses the unused value2 argument. Respawn clears the one-per-life key and marks player data dirty.

Validation: dotnet build VintageStory.slnx -c Release completed with 0 errors. make smoke reached WorldReady without fatal errors. The build reports four System.Drawing.Common vulnerability warnings.

Zaldaryon:
Trevor, the two remaining open conversations are covered by ae865be.

StratumKitGiver.DecodeStack reads new StackJson entries through JsonItemStack.Resolve and keeps Base64 for kits created by earlier builds. StratumCommandAccessCatalog includes the kit and kitedit entries, so the catalog exposes both configured privileges.

The other review points are resolved in the current diff. Please take another look.

trevorftp (approved):
Looks good

Issue #270, opened 2026-08-19
BlockSchematicStructure produces objects that are never collected by the GC


tehtelev:
### Stratum version

stratum-1.22.7-stratum.1-indev.1

### Server hardware

11700K, 64GB RAM

### Player count when problem occurs

0

### What you observe

When structures are generated in the world, they take up memory, and the garbage collector never frees them, even though the chunks have already been unloaded from memory.
I suspect this is also a vanilla bug. A PR for AnegoStudio should be duplicated then.

### Timings or profiler output

After active generation of chunks, we get the following picture of objects

<img width="1134" height="688" alt="Image" src="https://github.com/user-attachments/assets/b3f606bd-6de1-49d7-9bef-5819c680e5ef" />
<img width="1078" height="959" alt="Image" src="https://github.com/user-attachments/assets/7326e238-6249-4efa-9015-abe8854096f7" />
<img width="982" height="757" alt="Image" src="https://github.com/user-attachments/assets/c3fdb217-dbf6-45ee-a137-0d491d1fdfda" />
<img width="1118" height="902" alt="Image" src="https://github.com/user-attachments/assets/e9472722-e433-43d6-b94a-884b795f1f0d" />
<img width="1427" height="655" alt="Image" src="https://github.com/user-attachments/assets/3ae84627-08a0-4877-89b8-130b212f1601" />
<img width="1428" height="609" alt="Image" src="https://github.com/user-attachments/assets/1e1e0530-47c5-46be-885b-e374ea0d0067" />

Zaldaryon:
Opened #271 for this. Pulled the screenshots above and traced the retention chains directly: every one ends at `WorldGenStructuresConfig.LoadedSchematicsCache` via `GenStructures`, or the `GenDungeons` equivalent, at what looks like per-thread storage.

Turned out to be Stratum's own doing, not a vanilla issue to duplicate upstream: the concurrency patch to `GenStructures.cs` (the one that lets TerrainFeatures generate multiple columns at once) added `[ThreadStatic]` fields so the postpass can reuse a chunk's heightmap and climate data instead of recomputing it. That cached state only gets overwritten when the same worker thread happens to process another chunk's main pass, so once generation quiets down, whatever a thread last touched (the `GenStructures` instance and its schematic cache included) stays reachable from that thread indefinitely.

[...]

PR #271, opened 2026-08-19
Release GenStructures' cached column state after the postpass

Zaldaryon:
## Summary

Investigates #270 (BlockSchematicStructure objects never collected) by pulling and reading the actual profiler screenshots attached to the issue, then fixes the mechanism they point at.

Every retention chain in the screenshots ends the same way: `BlockSchematicStructure`/`BlockSchematicPartial` arrays reachable through `WorldGenStructuresConfig.LoadedSchematicsCache` via `GenStructures`, or through `GenDungeons` via a similar path, terminating at what looks like per-thread storage rather than a plain object reference. One screenshot names the exact field: `ServerEventManager.stratumCachedWorldSavedInvocations`.

That pointed at Stratum's own concurrency patch to `GenStructures.cs` (added to let TerrainFeatures generate multiple columns at once). It adds `[ThreadStatic]` fields (`stratumStateOwner`, `stratumHeightmap`, and a few siblings) so the postpass can reuse a chunk's already-computed heightmap and climate/forest values instead of recomputing them. The problem: that cached state is only ever overwritten when the same worker thread happens to process another chunk's main pass. Once worldgen quiets down, whatever a thread last touched (the owning `GenStructures` instance and its loaded schematic cache included) stays reachable from that thread for as long as the thread itself lives. Vanilla's own unpatched code doesn't have this property, since the equivalent state sits in plain instance fields tied to the `GenStructures` instance's own lifetime, not to a specific thread.

## Fix

Release `stratumStateOwner` and `stratumHeightmap` at the end of the postpass, the point at which a column's structure generation is actually finished. Both fields get set fresh by the main pass regardless of which branch runs, so clearing them right after the postpass reads them is always safe: the next chunk on that thread recomputes from scratch rather than reusing anything.

## Caveat

This is a real, demonstrable gap from vanilla's GC behavior, but a single heap snapshot can't distinguish "leaked forever" from "a worldgen thread that's mid-column right now." I can't reproduce a long-running server session with active generation followed by an idle period to directly confirm the counts stop growing. @tehtelev , since you filed this with the profiler already set up: if you get a chance to rerun the same capture after a while on this branch, that would close the loop for real.

## Type

- [x] Bug fix

## Checklist

- [x] `scripts/extract-patches.sh` ran clean. One file touched: `GenStructures.cs.patch`.
- [x] `dotnet build VintageStory.slnx -c Release` is green (0 errors, 0 new warnings).
- [x] Every vanilla edit has a `// Stratum` marker. The new lines carry one explaining why they're there.
- [x] Tested on a real server start, not just compilation (`make smoke`: PASS, reached `RunGame`).

## Related issues

Refs #270

### DIFF
--- a/patches/VSEssentials/Systems/WorldGen/Standard/ChunkGen/6.GenStructures/GenStructures.cs.patch
+++ b/patches/VSEssentials/Systems/WorldGen/Standard/ChunkGen/6.GenStructures/GenStructures.cs.patch
@@ -1,5 +1,5 @@
 diff --git a/VSEssentials/Systems/WorldGen/Standard/ChunkGen/6.GenStructures/GenStructures.cs b/VSEssentials/Systems/WorldGen/Standard/ChunkGen/6.GenStructures/GenStructures.cs
-index 95025db..7c475dd 100644
+index 95025db..a7465d4 100644
 --- a/VSEssentials/Systems/WorldGen/Standard/ChunkGen/6.GenStructures/GenStructures.cs
 +++ b/VSEssentials/Systems/WorldGen/Standard/ChunkGen/6.GenStructures/GenStructures.cs
 @@ -46,19 +46,44 @@ namespace Vintagestory.ServerMods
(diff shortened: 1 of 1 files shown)


tehtelev (approved):
It looks like everything is fine now.

PR #273, opened 2026-08-20
Fixing allocations inside ScheduleReadyTasks

tehtelev:
## Summary

In `TryClaimStage`, for stages with a neighbor gate, a 3×3 environment pin (`stratumPinnedColumns`) was set before the neighbor readiness check — for each claim attempt. Each pin/unpin is a mutation of `ConcurrentDictionary`, and each mutation allocates a new node. During pregen, the vast majority of claims are rejected by the gate (neighbors are still lagging behind), while the dispatcher and the `loadChunkAreaBlocking` wait loop repeatedly retry the same columns in a tight loop. The result: a "free" pin on the rejection path turned into an avalanche of short-lived dictionary nodes and CPU inside `ConcurrentDictionary` — even though a rejected claim doesn't start any stage and isn't required to protect residency at all.

**Bugfix.** The gate is split into two passes:

- **Pass 1** — full `ensurePrettyNeighbourhood` without a pin. Its side effects (requesting lagging neighbors, bumping `untilPass`) are preserved: they are what move the environment forward and are needed, among other things, in the failure path.

- The pin is set only if pass 1 has passed, that is, only when the stage is actually launched. Then **pass 2** — an authoritative re-gate under the pin: it closes the same race for which the pin existed (unloading can drop a neighbor between the residency check and holding). If the environment has changed, we fail and the pin is removed in `finally`.

The failure path now completely ignores `stratumPinnedColumns`; pin/unpin occurs once per column per stage transition, as intended. On the accepted path, pass 2 is mostly read-only: if the environment hasn't changed, `EnsureMinimumWorldgenPassAt` doesn't mutate anything. However, if a neighbor vanished in the gap between pass 1 and pass 2, pass 2 acts as a self-healing mechanism by re-adding and re-enqueuing the neighbor's request (the new comment in the diff documents this explicitly). The pin's job is not to make this call allocation-free, but to guarantee that when a neighbor IS still resident, nothing pulls it out from under the claim between the check and `PopulateChunk` actually reading it. The contract with the dispatcher and workers remains unchanged: on success, the pin is passed out along with the claim; on failure, both are released in `finally`.

## Type

- [ ] Bug fix
- [x] Performance
- [ ] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

## Performance numbers

**Before**
image
image
image

**After**
image
image

image

### DIFF
--- a/patches/VintagestoryLib/Vintagestory.Server/ServerSystemSupplyChunks.cs.patch
+++ b/patches/VintagestoryLib/Vintagestory.Server/ServerSystemSupplyChunks.cs.patch
@@ -1,8 +1,8 @@
 diff --git a/VintagestoryLib/Vintagestory.Server/ServerSystemSupplyChunks.cs b/VintagestoryLib/Vintagestory.Server/ServerSystemSupplyChunks.cs
-index 159b06c..ac6145a 100644
+index 159b06c..381605a 100644
 --- a/VintagestoryLib/Vintagestory.Server/ServerSystemSupplyChunks.cs
 +++ b/VintagestoryLib/Vintagestory.Server/ServerSystemSupplyChunks.cs
-@@ -1,139 +1,1094 @@
+@@ -1,139 +1,1109 @@
  using System;
  using System.Collections.Generic;
  using System.Linq;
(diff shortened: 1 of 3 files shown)


patches/VintagestoryLib/Vintagestory.Server/ServerSystemSupplyChunks.cs.patch
```diff
@@ -537,53 +537,46 @@ index 159b06c..ac6145a 100644
...
-+		// does not occur will always look like pure cost. Before rejecting a guard on
-+		// this path again, check whether the configuration under test has a gated
-+		// stage running concurrently, because that is the only regime where the queue
-+		// is under real pressure.
++		//   Only if pass 1 accepted do we pin, then run pass 2 as the authoritative
++		//   gate under the pin. Pass 2 closes the race the old pre-gate pin existed
++		//   for: between pass 1's residency observation and the hold, the unload
++		//   pass can drop a neighbour; pass 2 re-establishes residency while it is
++		//   held, and refuses (the finally releases the pin) if it can't. On this
++		//   path pass 2 is a read-only re-check: EnsureMinimumWorldgenPassAt
```
trevorftp: This isn't always read-only. If a neighbour actually unloads in the gap, this pass can create its request again.

trevorftp (approved):
No blockers from me, comment's are longer than I'd like but understandable for profiling history

patches/VintagestoryLib/Vintagestory.Server/ServerSystemSupplyChunks.cs.patch
```diff
@@ -600,11 +593,33 @@ index 159b06c..ac6145a 100644
...
 +			}
 +
-+			if (!ensurePrettyNeighbourhood(req))
++			if (claimedStage > StratumWorldGenStages.TerrainLate)
 +			{
-+				stage = -1;
-+				return false;
++				// Pass 1: unpinned filter + neighbour driver. Refusing here costs
++				// a few dictionary reads per neighbour and zero pin mutations.
++				if (!ensurePrettyNeighbourhood(req))
```
Zaldaryon, line 559: This moves the first `ensurePrettyNeighbourhood` call ahead of `StratumPinNeighbourhood`. On the first attempt, `req.prettified` is false, so this call can enter `EnsureQueueSpace`. That method pauses the workers and invokes `UnloadGeneratingChunkColumns`, which filters against `stratumPinnedColumns`. At this point the code has not pinned the request or its 3x3 neighbourhood, so cleanup can remove the request or one of its neighbours before pass 2. The later gate cannot restore the request that cleanup removed. Please keep the request protected through `EnsureQueueSpace`, or make that cleanup path respect the claim.

patches/VintagestoryLib/Vintagestory.Server/ServerSystemSupplyChunks.cs.patch
```diff
@@ -600,11 +593,33 @@ index 159b06c..ac6145a 100644
...
++				{
++					stage = -1;
++					return false;
++				}
++
++				// Pass 1 accepted: neighbours looked resident. "Looked" - the
++				// unload pass can drop one between the check and the hold, the
++				// exact check-then-act shape the pin exists to close. Pin, then
++				// re-check.
++				chunkthread.StratumPinNeighbourhood(req.chunkX, req.chunkZ);
```
Zaldaryon: This second gate does not make the pin authoritative against the normal unload path. `FindUnloadableChunkColumnCandidates` filters pins when it builds `mapChunkUnloadCandidates`, but `UnloadChunkColumns` consumes that list without a second pin check. The normal unload path can remove a candidate after pass 2 accepts and before `PopulateChunk` reads the neighbour. Please add the same handshake to `UnloadChunkColumns` or explain why this path cannot overlap.

Zaldaryon (changes requested):
I found two correctness issues in the new two-pass gate.

The first pass runs before the pin. On a first attempt, `ensurePrettyNeighbourhood` can call `EnsureQueueSpace`, which pauses the workers and invokes `UnloadGeneratingChunkColumns`. Since the code has not populated `stratumPinnedColumns`, that cleanup can remove the claimed request or a neighbour before the second pass. The later gate cannot restore the removed request.

The second pass depends on a pin check that the normal unload path does not use. `FindUnloadableChunkColumnCandidates` checks pins when it builds `mapChunkUnloadCandidates`, but `UnloadChunkColumns` does not check them before removal. A stale candidate can disappear after pass 2 accepts and before `PopulateChunk` reads the neighbour.

I bootstrapped the head with zero patch failures, built `VintageStory.slnx` with 0 errors, and ran the server smoke test to `WorldReady` without fatal errors. Those checks do not exercise the unload interleavings. The PR checks contain Discord notification jobs. Please add a deterministic pregen test with the exact seed, radius, queue size, worker count, and runtime versions, plus raw allocation and timing results. I would hold this PR until the unload handshake and measurement data are addressed.

patches/VintagestoryLib/Vintagestory.Server/ServerSystemSupplyChunks.cs.patch
Zaldaryon: Agreed. I traced this through `EnsureMinimumWorldgenPassAt`: when the neighbour disappears, the pass can `GetOrAdd` a `ChunkColumnLoadRequest`, update its target pass, and enqueue it. The second pass can mutate the request queue and allocate. Please remove the "read-only" and "no queue churn" claims, then include this interleaving in the benchmark.

Zaldaryon (approved):
Both issues from my last review are fixed.

The unpinned first call is handled: `EnsureQueueSpace` only runs while `req.prettified` is false, and the neighbourhood pin now goes up before that first call instead of after pass 1, so cleanup can no longer remove the request or a neighbour ahead of pass 2.

`UnloadChunkColumns` now has the other half of the handshake. `StratumTryMarkMapChunkDropping` commits to the removal, a pin taken before that point is caught by the immediate `StratumIsColumnPinned` recheck, and `EnsureMinimumWorldgenPassAt` refuses residency for any index the removal has already committed to. `StratumClearMapChunkDropping` runs in a finally, so a thrown exception during removal cannot leave the flag stuck. This closes the gap for Done columns that no longer carry a `ChunkColumnLoadRequest`.

One thing left over from this thread: the PR description still says pass 2 is read-only and adds no queue churn. That is no longer accurate on the path where a neighbour vanished and `EnsureMinimumWorldgenPassAt` re-adds and re-enqueues its request, which the new comment in the diff itself now says. Worth updating the description text, and a benchmark run that exercises that interleaving would make the perf claim solid, but neither blocks this landing in indev.

Approving.

tehtelev:
Yes. I just checked that pinned chunks that were subject to unloading are no longer unloaded. This is indeed true.
I'm updating the description and tests.

PR #275, opened 2026-08-21
Improvements to AVX instructions in ChunkDataLayer

tehtelev:
### What the optimization does

Previously, `CopyBlocksToUnsafe` and `CopyLightToUnsafe` decoded bitplanes → palette values ​​block by block, one after another. Now they do it 32 or 64 blocks at a time, using the processor's SIMD registers.

### Three paths with automatic selection

The code checks instruction support on the fly and selects the appropriate one:
- **Scalar** — fallback for older CPUs, works as before.
- **AVX2** — 32 blocks per iteration, for modern x86-64.
- **AVX-512** — 64 blocks per iteration, for even newer CPUs (Ice Lake+).

### How the fast path works

**1. Loading and classification.** We read words from the bitplanes directly through unsafe pointers. If all 32 bits in a word are the same, the word is called *uniform*. Uniforms don't require any unpacking: one `Vector256.Create(val)` and four unaligned entries, and you're done.

2. Mixed case (different blocks in a word).** This is where the real work happens:
- **Pack + Transpose.** The plane bits are decomposed into bytes using `vpshufb`, then an 8x8-bit butterfly transpose is applied, yielding the finished palette indices.
- **Lookup into palette.** For small palettes (≤16 elements), `vpermd` is used—one instruction fills 16 values ​​without accessing memory. For larger palettes, `gather` is used with a mask where invalid indices are replaced with zero (analogous to a bounds-check).

3. The downside is IsUniformWord.** When uncompressing (`SetBlocksBulkUnsafe`), AVX2 byte-compare checks in a single instruction whether 32 consecutive values ​​are equal. If so, the word is written with a single call to `Vector.Create`.

### What's Untouched (for now)

Scalar read paths (GetGeneralCase), lock logic, palette growth, and bitwise write operations—everything remains the same. AVX is used only in two methods: bulk block decoding and light decoding. Scalar fallback guarantees correct operation on any CPU.

## Type

- [ ] Bug fix
- [x] Performance
- [ ] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

## Performance numbers

Tested on the generation of 31417 chunk columns (6 threads).

The measurements are based on the execution time of ChunkData.CopyBlocksToUnsafe:
without instructions - 70778 ms;
AVX2 branch (now) - 2358 ms;
AVX2 branch (improved) - 1973 ms;
AVX512 branch (new) - 1836 ms.

But this is, of course, for my processor; the results may be different on other hardware.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -50,7 +50,7 @@ jobs:
         shell: pwsh
         env:
           STRATUM_VER: ${{ steps.ver.outputs.version }}
-        run: ./scripts/pack-release.ps1 -Rids linux-x64,win-x64 -OutDir release-out -Version "${{ steps.ver.outputs.version }}"
+        run: ./scripts/pack-release.ps1 -Rids linux-x64,linux-arm64,win-x64 -OutDir release-out -Version "${{ steps.ver.outputs.version }}"
 
       - name: Upload release artifacts
         uses: softprops/action-gh-release@v2
--- a/StratumServer/StratumServer.csproj
+++ b/StratumServer/StratumServer.csproj
@@ -4,7 +4,10 @@
     <GenerateAssemblyInfo>False</GenerateAssemblyInfo>
     <OutputType>Exe</OutputType>
     <TargetFramework>$(FrameworkVersion)</TargetFramework>
-    <PlatformTarget>x64</PlatformTarget>
+    <!-- RID-less builds stay x64. When a RuntimeIdentifier is supplied (pack-release.ps1 publishes
+         per-RID), let the SDK infer the platform from it - an explicit x64 here fails NETSDK1032
+         against -r linux-arm64. -->
+    <PlatformTarget Condition="'$(RuntimeIdentifier)' == ''">x64</PlatformTarget>
     <RootNamespace>StratumServer</RootNamespace>
     <Configurations>Debug;Release;PerfTest</Configurations>
     <EmbedPatchedFiles Condition="'$(EmbedPatchedFiles)' == ''">false</EmbedPatchedFiles>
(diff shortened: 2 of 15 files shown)


Zaldaryon (approved):
Applied this patch onto the vanilla 1.22.7 baseline and pulled the resulting ChunkDataLayer.cs into a standalone harness rather than reviewing the diff on its own. Full round-trip battery, 33 fixtures (all-air, uniform, low and high cardinality, near and over capacity, three palette-growth-timing variants, 20 randomized realistic-terrain fixtures), run twice: natively (this machine has AVX2, not AVX-512) and again with DOTNET_EnableAVX2=0 forcing the scalar fallback. Diffing the two runs, SHA-256 hashes included, is empty except the Avx2.IsSupported header line. 31 of 33 fixtures round-trip identically between the two paths; the remaining two hit the same pre-existing vanilla capacity-guard exception #266's review already documented, on both paths, and this diff does not touch the write side (SetBlocksBulkUnsafe/PaletteIndexForBulk) at all, so that finding is unrelated to this PR.

The rewritten AVX2 path is also faster than the one it replaces: an isolated decode microbenchmark on this machine measured 13.28x over the old per-position Get() loop, up from the 4.5x-5.55x #266's own AVX2 path got. Consistent with the performance table in the description.

One gap I want to be direct about: the AVX-512 path cannot be exercised here. This machine's CPU has no AVX-512 support (confirmed via lscpu), and unlike AVX2 there is no environment variable that can force a code path to run on hardware that structurally lacks the instructions, only ones that can force it off where it exists. I read DecodePlanesAvx512 closely instead: it shares LoadWordsAndClassify2's word-loading and uniform-word fast path with the now-verified AVX2 code, the vpermd (Vector512.Shuffle) lookup for palettes <= 16 and the masked Avx2.GatherMaskVector256 pair for larger ones both preserve the same idx < palLen ? palette[idx] : 0 bounds-safety the verified paths already use, and the trailing single-word tail branch is unreachable dead code (the loop advances by 2 and 1024 is even) but harmless. This is a real, unclosed verification gap in the same category the project already accepted for linux-arm64 after #266: it needs Ice-Lake-or-newer x86-64 hardware to actually confirm, not more reading.

Linux bootstrap and build are clean: scripts/bootstrap.sh applied every patch with zero failures, dotnet build VintageStory.slnx -c Release gave zero errors. The Linux smoke test itself crashed on this build, but I reproduced the identical crash (same exception, same stack trace) on a clean unpatched indev rebuild too, so it's a pre-existing gap in this workspace's own smoke-test setup against 1.22.7, unrelated to this diff. Filed as StratumServer/Stratum#276 rather than held against this PR.

Approving on the strength of the round-trip and benchmark evidence above. Worth flagging in this PR directly, the same way linux-arm64 confirmation was flagged after #266, that AVX-512 correctness on real hardware is still owed before anyone should fully trust that path in production.

Issue #276, opened 2026-08-22
Linux smoke test crashes on 1.22.7: Mono.Cecil cannot resolve VintagestoryAPI for VSCreativeMod and VSEssentials


Zaldaryon:
### Stratum version

`indev` at HEAD, clean checkout. Not tied to a pending pull request.

### Base game version

1.22.7

### OS and .NET runtime

Linux and WSL, .NET 10

### Stack trace

```
[Server Error] [VSCreativeMod.dll] An exception was thrown trying to load the ModInfo:
[Server Error] [VSCreativeMod.dll] Exception: Failed to resolve assembly: 'VintagestoryAPI, Version=1.22.7.0, Culture=neutral, PublicKeyToken=null'
   at Mono.Cecil.BaseAssemblyResolver.Resolve(AssemblyNameReference name, ReaderParameters parameters)
   ...
   at Vintagestory.Common.ModContainer.LoadModInfoFromCode(...)
```

The same failure hits `VSEssentials.dll`. With both unresolved, `survival`'s `game` dependency is reported missing, no mod systems load, and the server then throws an unhandled `NullReferenceException` in `WorldConfig.loadWorldConfigValuesFromPlaystyle()`, called from `SaveGame.SetNewWorldConfig` during `ServerSystemModHandler.OnLoadAssets`, which kills the process before it reaches `RunGame`.

### What was happening

The server was loading assets during `scripts/smoke-test.sh` on a clean `indev` build. `VintagestoryAPI.dll` is present in the build output next to the mod DLLs, so this is Mono.Cecil's own resolver, used only for the `IsModSystem` metadata scan at `ModContainer.cs:524`, not being given the search path the normal .NET loader uses. It is not a missing file.

### Reproduction steps

```
bash scripts/bootstrap.sh
dotnet build VintageStory.slnx -c Release
bash scripts/smoke-test.sh
```

On a clean `indev` checkout, Linux or WSL, .NET 10, base version 1.22.7.

### Mods loaded

Stock bundled mods only: `VSCreativeMod`, `VSEssentials`, `VSSurvivalMod`. No third-party mods.

### Note

This blocks the smoke-test half of the pre-PR gate in `docs/agent/stratum.md` and `CONTRIBUTING.md` for every branch, not one pull request. Likely fix area: give the Cecil resolver the mod output directory as an extra search path in the `IsModSystem` scan.

Zaldaryon:
[...]

PR #279, opened 2026-08-22
GenTerra.stratumGenerate optimizations

tehtelev:
## Summary

1. Lock-free merging of thread-local layer buffers.
Removed mergeLock from the localFinally callback in stratumGenerate. Merging is idempotent (only true→false transitions), and reading layerFullySolid/layerFullyEmpty occurs strictly after Parallel.For completes — the lock only protected itself and created contention at the end of the loop.

2. Compact list of non-zero landforms in a hot threshold loop.
Instead of scanning all landforms for each block (weight == 0 → continue + jagged access terrainYThresholds[i]), the column builds a compact list of nonzero landforms once with premultiplied weights wA = w·(1−slide) and wB = w·slide (a consequence of Lerp's end-to-end linearity: Σ wᵢ Lerp(aᵢ,bᵢ,t) = Σ aᵢ wᵢ(1−t) + bᵢ wᵢ t). The hot loop consists of K≈2–4 iterations of "2 loads + 2 FMAs", with early exit by noiseBoundMax preserved. Buffers (CompactWeightsA/B, CompactThresholds) live in ThreadLocalTempData — no per-column allocations.
Note: An intermediate version with a full precomputed threshold profile for the entire height was rejected — it destroyed the benefit of an early break and introduced a regression.

3. Dedicated shared thread pool for the internal Parallel.For (StratumColumnScheduler).
New TaskScheduler with maxThreads dedicated background threads, shared across all concurrent calls to stratumGenerate; MaxDegreeOfParallelism is removed — only the scheduler limits it. Solves three ThreadPool issues in short burst loops: warmup "laddering" (threads weren't created in time during the For loop), contention of the global ThreadPool queue (Monitor.Enter_SlowPath), and idle tails of one request while workers for other requests are free (the task queue is now shared). The vanilla generate() body is untouched — mods with Harmony patches get the previous behavior. Scheduler lifecycle is hardened: it is reused across `/wgen regen` (if the thread budget is unchanged) to avoid unnecessary thread churn, and `QueueTask` safely falls back to inline execution if the scheduler is disposed, preventing deadlocks during shutdown.

4. Safe nested column generation via per-thread frame stack.
OnChunkColumnGen now pushes a StratumColumnBufferFrame onto a [ThreadStatic] Stack<StratumColumnBufferFrame> and saves/restores the previous frame's state on exit, making nested calls (e.g., from mods that re-enter worldgen) correct. The entire vanilla `generate()` path — including the pre-publish and post-restore of shared instance fields — is serialized with `stratumVanillaGenerateLock` so concurrent columns and nested re-entrant calls don't stomp each other's state (taperMap, columnResults, layerFullySolid/Empty).

5. Call-local border indices and double-checked landform init.
borderIndicesByCardinal is now allocated per call (cheap — 8 ints) rather than shared, eliminating a hidden race in the smoothing block. Landforms initialization uses proper double-checked locking under stratumLandformsInitLock.

## Type

- [ ] Bug fix
- [x] Performance
- [ ] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

## Performance numbers

Tested on the generation of 31417 chunk columns (6 threads).

The cycle time may vary from measurement to measurement, so you need to compare the ratio of your own time to the total time:
before - 79.8%;
after - 66.9 %.

This means that we can say that the internal work has been reduced, while the external method calls have not been touched.

**Before**
image
image
image

**After**
image
image
image

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -50,7 +50,7 @@ jobs:
         shell: pwsh
         env:
           STRATUM_VER: ${{ steps.ver.outputs.version }}
-        run: ./scripts/pack-release.ps1 -Rids linux-x64,win-x64 -OutDir release-out -Version "${{ steps.ver.outputs.version }}"
+        run: ./scripts/pack-release.ps1 -Rids linux-x64,linux-arm64,win-x64 -OutDir release-out -Version "${{ steps.ver.outputs.version }}"
 
       - name: Upload release artifacts
         uses: softprops/action-gh-release@v2
--- a/StratumServer/StratumServer.csproj
+++ b/StratumServer/StratumServer.csproj
@@ -4,7 +4,10 @@
     <GenerateAssemblyInfo>False</GenerateAssemblyInfo>
     <OutputType>Exe</OutputType>
     <TargetFramework>$(FrameworkVersion)</TargetFramework>
-    <PlatformTarget>x64</PlatformTarget>
+    <!-- RID-less builds stay x64. When a RuntimeIdentifier is supplied (pack-release.ps1 publishes
+         per-RID), let the SDK infer the platform from it - an explicit x64 here fails NETSDK1032
+         against -r linux-arm64. -->
+    <PlatformTarget Condition="'$(RuntimeIdentifier)' == ''">x64</PlatformTarget>
     <RootNamespace>StratumServer</RootNamespace>
     <Configurations>Debug;Release;PerfTest</Configurations>
     <EmbedPatchedFiles Condition="'$(EmbedPatchedFiles)' == ''">false</EmbedPatchedFiles>
@@ -21,6 +24,9 @@
 
   <ItemGroup>
     <Compile Include="$(StratumInfoPath)" Link="StratumInfo.cs" Condition="'$(StratumInfoPath)' != ''" />
+    <!-- Pinned arm64 native overlay releases (see forks.json's arm64NativeOverlays), read back at
+         runtime by VanillaBootstrap so arm64 asset resolution needs no GitHub API call. -->
+    <EmbeddedResource Include="..\forks.json" LogicalName="Stratum.forks.json" />
   </ItemGroup>
 
   <Target Name="ValidateStratumInfoPath" BeforeTargets="CoreCompile">
--- a/StratumServer/VanillaBootstrap.cs
+++ b/StratumServer/VanillaBootstrap.cs
@@ -14,6 +14,17 @@ internal static class VanillaBootstrap
 {
 	private const string VersionManifestUrl = "https://api.vintagestory.at/stable-unstable.json";
 
+	// Anego publishes the aarch64 natives as a standalone overlay; the main manifest is x64-only.
+	// Pinned in forks.json (arm64NativeOverlays), embedded into the assembly - resolving it needs
+	// no GitHub API call at runtime.
+	private const string Arm64CatalogResourceName = "Stratum.forks.json";
+	private static readonly string[] RequiredArm64NativeLibraries =
+	{
+		"libe_sqlite3.so",
+		"libSkiaSharp.so",
+		"libzstd.so"
+	};
+
 	internal static void EnsureVanillaAssets(string baseGameVersion, bool refresh)
 	{
 		string installDir = AppContext.BaseDirectory;
@@ -22,9 +33,25 @@ internal static class VanillaBootstrap
 		bool markerExists = File.Exists(markerPath);
 		string currentMarker = markerExists ? File.ReadAllText(markerPath).Trim() : string.Empty;
 		bool markerMatches = markerExists && currentMarker == expectedMarker;
+		bool isLinuxArm64 = IsLinuxArm64();
+		Arm64Asset arm64Asset = default;
+		if (isLinuxArm64)
+		{
+			// Resolve from the embedded catalog before the base-marker fast path. An install
+			// created on x64 still needs the arm64 overlay when it moves to an arm64 host.
+			arm64Asset = ResolveArm64Asset(baseGameVersion);
+		}
 
 		if (!refresh && markerMatches)
 		{
+			if (!isLinuxArm64)
+			{
+				return;
+			}
+
+			string existingCacheDir = Path.Combine(installDir, ".vanilla-cache");
+			Directory.CreateDirectory(existingCacheDir);
+			ApplyArm64NativeOverlay(installDir, existingCacheDir, baseGameVersion, arm64Asset, refresh);
 			return;
 		}
 
@@ -90,6 +117,13 @@ internal static class VanillaBootstrap
 			Console.WriteLine($"Stratum: installed {copied} vanilla file(s) (existing files were preserved)");
 		}
 
+		// Anego ships no linux-arm64 server archive, so the files just laid down are x86-64.
+		// Swap the handful of architecture-specific natives before PatchedFileOverlay runs.
+		if (isLinuxArm64)
+		{
+			ApplyArm64NativeOverlay(installDir, cacheDir, baseGameVersion, arm64Asset, refresh);
+		}
+
 		File.WriteAllText(markerPath, expectedMarker);
 	}
 
@@ -134,6 +168,227 @@ internal static class VanillaBootstrap
 		}
 
 		return new ArchiveInfo(fileName, url, md5, isZip);
+	}
+
+	// Replaces the x86-64 natives from the official archive with the aarch64 builds Anego publishes
+	// as a separate overlay release. Only *.so files are taken: everything else in that overlay is
+	// either portable IL or VintagestoryServer host files, and the overlay lags the base game by a
+	// patch release or two, so copying its managed/host files would risk a version mismatch against
+	// Stratum's patched assemblies. The natives are self-contained third-party libraries
+	// (SkiaSharp, e_sqlite3, zstd) whose ABI does not track Vintage Story patch releases.
+	private static void ApplyArm64NativeOverlay(string installDir, string cacheDir, string baseGameVersion, Arm64Asset asset, bool refresh)
+	{
+		string markerPath = Path.Combine(installDir, ".stratum-arm64-natives");
+		string expectedMarker = BuildArm64Marker(baseGameVersion, asset);
+		if (!refresh && File.Exists(markerPath) && File.ReadAllText(markerPath).Trim() == expectedMarker)
+		{
+			try
+			{
+				ValidateArm64NativeLibraries(installDir);
+				Console.WriteLine("Stratum: arm64 natives already in place for " + baseGameVersion + "; leaving them untouched");
+				return;
+			}
+			catch (InvalidOperationException exception)
+			{
+				Console.WriteLine("Stratum: arm64 native layout is stale; reapplying the overlay: " + exception.Message);
+			}
+		}
+
+		Console.WriteLine($"Stratum: linux-arm64 host detected; fetching native overlay {asset.Name}");
+
+		string archivePath = Path.Combine(cacheDir, asset.Name);
+		if (File.Exists(archivePath) && !VerifyArm64Digest(archivePath, asset.Digest))
+		{
+			Console.WriteLine($"Stratum: cached {asset.Name} failed checksum; downloading a fresh copy");
+			File.Delete(archivePath);
+		}
+
+		if (!File.Exists(archivePath))
+		{
+			DownloadFile(asset.Url, archivePath);
+			if (!VerifyArm64Digest(archivePath, asset.Digest))
+			{
+				File.Delete(archivePath);
+				throw new InvalidOperationException("Downloaded arm64 native overlay failed SHA256 verification: " + asset.Name);
+			}
+		}
+		else
+		{
+			Console.WriteLine($"Stratum: using cached {archivePath}");
+		}
+
+		string extractDir = Path.Combine(cacheDir, "extract-arm64");
+		if (Directory.Exists(extractDir))
+		{
+			Directory.Delete(extractDir, recursive: true);
+		}
+		Directory.CreateDirectory(extractDir);
+
+		Console.WriteLine($"Stratum: unpacking {asset.Name}");
+		ExtractTarGz(archivePath, extractDir);
+
+		string sourceRoot = FindContentRoot(extractDir);
+		int copied = 0;
+		foreach (string sourcePath in Directory.EnumerateFiles(sourceRoot, "*", SearchOption.AllDirectories))
+		{
+			if (!IsNativeLibrary(sourcePath))
+			{
+				continue;
+			}
+
+			string rel = Path.GetRelativePath(sourceRoot, sourcePath);
+			string destPath = Path.Combine(installDir, rel);
+			string destDir = Path.GetDirectoryName(destPath);
+			if (destDir != null)
+			{
+				Directory.CreateDirectory(destDir);
+			}
+			File.Copy(sourcePath, destPath, overwrite: true);
+			copied++;
+		}
+
+		Directory.Delete(extractDir, recursive: true);
+
+		if (copied == 0)
+		{
+			throw new InvalidOperationException("arm64 native overlay " + asset.Name + " contained no shared libraries; refusing to run against x86-64 natives.");
+		}
+
+		RemoveUnusedArm64NativeLibraries(installDir);
+		ValidateArm64NativeLibraries(installDir);
+		Console.WriteLine($"Stratum: replaced {copied} native librar(ies) with aarch64 builds from {asset.Name}");
+		File.WriteAllText(markerPath, expectedMarker);
+	}
+
+	private static bool IsLinuxArm64()
+	{
+		return RuntimeInformation.IsOSPlatform(OSPlatform.Linux)
+			&& RuntimeInformation.ProcessArchitecture == Architecture.Arm64;
+	}
+
+	private static string BuildArm64Marker(string baseGameVersion, Arm64Asset asset)
+	{
+		return "version=" + baseGameVersion + "\nasset=" + asset.Name + "\ndigest=" + asset.Digest;
+	}
+
+	private static void RemoveUnusedArm64NativeLibraries(string installDir)
+	{
+		// The official x64 archive includes OpenAL for the client, but the dedicated server
+		// never loads it and the archive has no arm64 replacement. Do not leave an x64 ELF
+		// in an arm64 install where a mod or future code could load it accidentally.
+		string openAlPath = Path.Combine(installDir, "Lib", "libopenal.so.1");
+		if (File.Exists(openAlPath))
+		{
+			File.Delete(openAlPath);
+			Console.WriteLine("Stratum: removed unused x86-64 libopenal.so.1 from arm64 install");
+		}
+	}
+
+	private static void ValidateArm64NativeLibraries(string installDir)
+	{
+		string libDir = Path.Combine(installDir, "Lib");
+		if (!Directory.Exists(libDir))
+		{
+			throw new InvalidOperationException("arm64 native directory is missing: Lib");
+		}
+
+		foreach (string name in RequiredArm64NativeLibraries)
+		{
+			string path = Path.Combine(libDir, name);
+			if (!File.Exists(path))
+			{
+				throw new InvalidOperationException("required arm64 native library is missing: Lib/" + name);
+			}
+		}
+
+		foreach (string path in Directory.EnumerateFiles(libDir, "*", SearchOption.TopDirectoryOnly))
+		{
+			if (!IsNativeLibrary(path))
+			{
+				continue;
+			}
+
+			if (!IsArm64Elf(path))
+			{
+				throw new InvalidOperationException("native library is not an AArch64 ELF: " + Path.GetRelativePath(installDir, path));
+			}
+		}
+	}
+
+	private static bool IsArm64Elf(string path)
+	{
+		byte[] header = new byte[20];
+		using FileStream stream = File.OpenRead(path);
+		stream.ReadExactly(header);
+		if (header[0] != 0x7f || header[1] != (byte)'E' || header[2] != (byte)'L' || header[3] != (byte)'F')
+		{
+			return false;
+		}
+
+		// Linux arm64 uses ELF64 little-endian and e_machine EM_AARCH64 (183).
+		return header[4] == 2 && header[5] == 1 && header[18] == 183 && header[19] == 0;
+	}
+
+	private static bool IsNativeLibrary(string path)
+	{
+		string fileName = Path.GetFileName(path);
+		return fileName.EndsWith(".so", StringComparison.OrdinalIgnoreCase)
+			|| fileName.Contains(".so.", StringComparison.OrdinalIgnoreCase);
+	}
+
+	// Overlay releases are cut per minor version, so a 1.22.6 base resolves the pinned 1.22 entry.
+	// Pinned in forks.json rather than resolved from the GitHub API at runtime: it keeps every
+	// arm64 asset URL and digest reviewable in source control, and needs no network round-trip
+	// (or GITHUB_TOKEN, to dodge the anonymous rate limit) just to boot.
+	private static Arm64Asset ResolveArm64Asset(string baseGameVersion)
+	{
+		string majorMinor = MajorMinor(baseGameVersion);
+
+		using Stream resource = typeof(VanillaBootstrap).Assembly.GetManifestResourceStream(Arm64CatalogResourceName)
+			?? throw new InvalidOperationException("Embedded resource missing: " + Arm64CatalogResourceName);
+		using JsonDocument document = JsonDocument.Parse(resource);
+
+		if (document.RootElement.TryGetProperty("arm64NativeOverlays", out JsonElement overlays))
+		{
+			foreach (JsonElement entry in overlays.EnumerateArray())
+			{
+				if (!string.Equals(RequiredString(entry, "version"), majorMinor, StringComparison.OrdinalIgnoreCase))
+				{
+					continue;
+				}
+
+				string name = RequiredString(entry, "name");
+				string url = RequiredString(entry, "url");
+				string sha256 = RequiredString(entry, "sha256");
+				return new Arm64Asset(name, url, "sha256:" + sha256);
+			}
+		}
+
+		throw new InvalidOperationException(
+			"No pinned arm64 native overlay for Vintage Story " + majorMinor + ".x in forks.json. " +
+			"Add one under arm64NativeOverlays from https://github.com/anegostudios/VintagestoryServerArm64/releases.");
+	}
+
+	private static string MajorMinor(string version)
+	{
+		string[] parts = version.Split('.');
+		return parts.Length >= 2 ? parts[0] + "." + parts[1] : version;
+	}
+
+	private static bool VerifyArm64Digest(string path, string digest)
+	{
+		// digest is sourced from our own pinned forks.json, not a remote response, so a missing or
+		// malformed value is a config mistake - fail closed rather than silently trusting the file.
+		const string Sha256Prefix = "sha256:";
+		if (string.IsNullOrWhiteSpace(digest) || !digest.StartsWith(Sha256Prefix, StringComparison.OrdinalIgnoreCase))
+		{
+			throw new InvalidOperationException($"arm64 native overlay entry in forks.json has no usable sha256 digest: '{digest}'");
+		}
+
+		using FileStream stream = File.OpenRead(path);
+		using SHA256 sha = SHA256.Create();
+		string actual = Convert.ToHexString(sha.ComputeHash(stream));
+		return string.Equals(actual, digest.Substring(Sha256Prefix.Length), StringComparison.OrdinalIgnoreCase);
 	}
 
 	private static void DownloadFile(string url, string destination)
@@ -231,13 +486,13 @@ internal static class VanillaBootstrap
 	{
 		if (!element.TryGetProperty(propertyName, out JsonElement property))
 		{
-			throw new InvalidOperationException("Anego manifest entry is missing property: " + propertyName);
+			throw new InvalidOperationException("JSON entry is missing property: " + propertyName);
 		}
 
 		string value = property.GetString();
 		if (string.IsNullOrWhiteSpace(value))
 		{
-			throw new InvalidOperationException("Anego manifest entry has empty property: " + propertyName);
+			throw new InvalidOperationException("JSON entry has empty property: " + propertyName);
 		}
 
 		return value;
@@ -265,5 +520,19 @@ internal static class VanillaBootstrap
 		public string Url { get; }
 		public string Md5 { get; }
 		public bool IsZip { get; }
+	}
+
+	private readonly struct Arm64Asset
+	{
+		public Arm64Asset(string name, string url, string digest)
+		{
+			Name = name;
+			Url = url;
+			Digest = digest;
+		}
+
+		public string Name { get; }
+		public string Url { get; }
+		public string Digest { get; }
 	}
 }
--- a/docs/BUILDING.md
+++ b/docs/BUILDING.md
@@ -78,10 +78,34 @@ Commit the updated `patches/` and `sources/` files.
 ## Release Zips
 
 ```powershell
-.\scripts\pack-release.ps1 -Rids win-x64,linux-x64 -OutDir release-out
+.\scripts\pack-release.ps1 -Rids win-x64,linux-x64,linux-arm64 -OutDir release-out
 ```
 
 Release zips contain `StratumServer` plus Stratum patched managed files. They do
 not contain the full official Vintage Story server archive or files. On first run, the
 launcher downloads and verifies the official archive, extracts it, writes the
 patched files, and then starts the server.
+
+Every project except `StratumServer` builds AnyCPU, and the launcher is published
+framework-dependent, so the only architecture-specific artifact is the apphost. The SDK
+cross-targets it from an x64 host - `linux-arm64` needs no arm64 build machine.
+
+## arm64
+
+Anego publishes no `linux-arm64` server archive, so on an arm64 host the launcher still
+downloads the official `linux-x64` archive for its assets and IL, then replaces the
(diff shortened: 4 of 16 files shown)


Zaldaryon (changes requested):
I reviewed head `04ab77ff73650d7d0d5844e95b3af717b7763810` and found three correctness blockers.

1. The rented per-column buffers are still published through shared instance fields. `stratumPublishColumnBuffers` assigns `taperMap`, `columnResults`, `layerFullySolid`, and `layerFullyEmpty`, while `OnChunkColumnGen` assigns `borderIndicesByCardinal`. These are instance fields, not thread-local state. The existing Stratum Terrain stage is raised to the worker count, so two overlapping columns on the same `GenTerra` instance can publish different buffers and then read or write the other column's arrays. This contradicts the claim that concurrent columns have isolated state. Please use per-worker generator instances or pass the active buffer explicitly through every path, including the compatibility path, and add an overlap test that verifies each column keeps its own buffers.

2. The ocean-map bottom samples use the climate-map scale. In both generated methods, `ofac` comes from `oceanMap.InnerSize`, while `cfac` comes from `climateMap.InnerSize`. The patch changes both bottom ocean coordinates from `ofac` to `cfac` at patch lines 438-439 and 587-588. The base generated source uses `ofac` for all four ocean samples. This changes terrain when the map resolutions differ and also contradicts the description that the vanilla body is untouched. Please leave the vanilla hunk unchanged, use `ofac` for every `oceanMap` coordinate in the optimized path, and add a parity test with distinct climate and ocean map resolutions.

3. The nested buffer frame is not restored. The cleanup calls `frames.Pop()` and returns the inner buffer, but does not republish the previous frame after the pop. If a nested same-thread call occurs, the outer generator resumes with the inner buffer still published through the instance fields. The comment explicitly says nested calls are supported. Please restore the previous frame after popping, or remove nested-call support and add a regression test for the supported behavior.

The patch bootstraps cleanly and the Release build completes with 0 errors. The local smoke test cannot reach RunGame because the current Linux 1.22.7 baseline hits the known Mono.Cecil resolver failure tracked in [issue #276](https://github.com/StratumServer/Stratum/issues/276). GitHub currently reports no status checks or workflow runs for this head, and there are no existing reviews, comments, or threads. The performance section has screenshots and prose but no reproducible command, seed, configuration table, or parity output. Please address the correctness blockers and add reproducible concurrent parity and performance evidence before approval.

Zaldaryon:
The existing `CHANGES_REQUESTED` review remains valid. I found one additional correctness blocker in the scheduler. `GenTerra.cs.patch` lines 992-995 implement `TryExecuteTaskInline` for non queued tasks, so a caller of `stratumGenerate` can execute a task outside the `maxThreads` worker array. A .NET 10 harness using a limit of two and four concurrent callers observed six active bodies. That makes `MaximumConcurrencyLevel` an inaccurate global cap and can oversubscribe the reserved CPU budget. Please disable external inline execution or account for caller execution in one global cap, then add a regression that asserts active bodies never exceed the configured limit and measures tick latency. An isolated smoke run reached `WorldReady`, but another clean run hit issue #276. Document the command and environment before treating smoke as a gate.

tehtelev:
What has been done now:

- Fix ocean map bottom samples using climate map scale (cfac instead of ac) 
in both generate() and stratumGenerate(). This caused terrain differences 
when climate and ocean maps had different resolutions.

- Fix nested OnChunkColumnGen buffer frame restoration. After Pop, the previous 
frame's buffers and border indices are now restored to instance fields, making 
the stack a proper snapshot/restore mechanism for nested calls from mods.

- Disable inline task execution in StratumColumnScheduler. TryExecuteTaskInline 
now returns false for all tasks, enforcing the global cap at MaximumConcurrencyLevel. 
All tasks execute only on worker threads; the calling worldgen thread waits via 
Task.Wait rather than participating in work. This prevents oversubscription 
the reserved CPU budget when multiple worldgen threads call stratumGenerate 
concurrently.

Trade-off: Caller latency is slightly higher (waits instead of helping), but this
guarantees active bodies never exceed maxThreads, preserving the reserved CPU budget.

tehtelev:
Thanks for the detailed review — all items are addressed.

**1. Scheduler cap.** `TryExecuteTaskInline` now unconditionally returns `false`. All tasks are queued and execute only on the `maxThreads` worker array, so `MaximumConcurrencyLevel` is the exact global cap; the calling thread joins via `Task.Wait` instead of running bodies. The latency-vs-cap trade-off is documented in a code comment.

**2. Regression harness.** The repo has no unit-test project and the scheduler can only be exercised inside the game process, so the regression test is an in-game console command running against the real scheduler: `/wgen testscheduler <maxThreads> <taskCount>` (server side, requires `controlserver`). It creates a fresh `StratumColumnScheduler`, submits `taskCount` 100 ms bodies through `Task.Factory.StartNew(..., scheduler)`, tracks concurrent active bodies via `Interlocked`, waits for completion, and prints observed max concurrency, total time and per-task latency. Assertions: `maxActive ≤ maxThreads` (cap) and `maxActive == min(taskCount, maxThreads)` (no starvation). The harness is sensitive to exactly the reported regression class: with the old inline behavior a caller executing bodies outside the worker array would push observed concurrency above the cap.
Command

```
                .BeginSubCommand("testscheduler")
                .WithDescription("Test StratumColumnScheduler concurrency cap")
                .WithArgs(parsers.OptionalInt("maxThreads", 2), parsers.OptionalInt("taskCount", 4))
                .HandleWith(OnCmdTestScheduler)
                .EndSubCommand()
```

Called code
```
      private TextCommandResult OnCmdTestScheduler(TextCommandCallingArgs args)
        {
            int maxThreads = args[0] is int mt ? mt : 2;
            int taskCount = args[1] is int tc ? tc : 4;
            var scheduler = new StratumColumnScheduler(maxThreads);
            int activeBodies = 0;
            int maxActiveBodies = 0;
            object lockObj = new object();
            var stopwatch = System.Diagnostics.Stopwatch.StartNew();

            var tasks = new Task[taskCount];
            for (int i = 0; i < taskCount; i++)
            {
                tasks[i] = Task.Factory.StartNew(() =>
                {
                    int current = Interlocked.Increment(ref activeBodies);

                    lock (lockObj)
                    {
                        maxActiveBodies = Math.Max(maxActiveBodies, current);
                    }

                    Thread.Sleep(100);

                    Interlocked.Decrement(ref activeBodies);
                }, CancellationToken.None, TaskCreationOptions.None, scheduler);
            }

            Task.WaitAll(tasks);
            stopwatch.Stop();

            scheduler.Dispose();

            // CORRECTED: Cap is respected if maxActiveBodies <= maxThreads
            // AND we actually used available resources: maxActiveBodies == min(taskCount, maxThreads)
            int expectedMax = Math.Min(taskCount, maxThreads);
            bool capRespected = maxActiveBodies <= maxThreads;
            bool resourcesUsed = maxActiveBodies == expectedMax;
            bool passed = capRespected && resourcesUsed;

            double avgLatency = (double)stopwatch.ElapsedMilliseconds / taskCount;

            string result = $"[StratumColumnScheduler Test]\n" +
                           $"Configuration: {maxThreads} max threads, {taskCount} concurrent tasks\n" +
                           $"Max active bodies observed: {maxActiveBodies}\n" +
                           $"Expected max: {expectedMax}\n" +
                           $"Cap respected (≤ {maxThreads}): {(capRespected ? "YES" : "NO")}\n" +
                           $"Resources used (== {expectedMax}): {(resourcesUsed ? "YES" : "NO")}\n" +
                           $"Total time: {stopwatch.ElapsedMilliseconds}ms\n" +
                           $"Avg latency per task: {avgLatency:F2}ms\n" +
                           $"Result: {(passed ? "PASS ✓" : "FAIL ✗")}";

            if (args.Caller.Player is IServerPlayer player)
            {
                player.SendMessage(args.Caller.FromChatGroupId, result,
                    passed ? EnumChatType.CommandSuccess : EnumChatType.CommandError);
            }
            else
            {
                api.Logger.Notification(result);
            }

            return TextCommandResult.Success(passed ? "Scheduler test passed" : "Scheduler test FAILED");
        }

```

**3. Results** (Windows x64, 16 logical cores, dedicated server):

| maxThreads | tasks | max active | total ms | avg ms/task | verdict |
|---|---|---|---|---|---|
| 2 | 4 | 2 | 201 | 50.25 | cap ok |
| 2 | 8 | 2 | 402 | 50.25 | cap ok |
| 4 | 2 | 2 | 100 | 50.00 | cap ok |
| 6 | 6 | 6 | 101 | 16.83 | cap ok |
| 6 | 3 | 3 | 100 | 33.33 | cap ok |
| 6 | 100 | 6 | 1709 | 17.09 | cap ok |
| 6 | 1000 | 6 | 16791 | 16.79 | cap ok |

The cap was never violated in any configuration. Total time matches the theoretical minimum `ceil(tasks/threads)·100 ms` within ~1%, i.e. workers are fully saturated and the queue adds no measurable serialization; per-task latency scales as `100 ms / threads`, which is the requested tick-latency evidence.

The `Server may be overloaded. A tick took Nms` warnings during the 100/1000-task runs are a harness artifact: the command handler blocks the main thread in `Task.WaitAll` for the whole test. In production `stratumGenerate` runs on worldgen threads and never blocks the main tick.

tehtelev:
**Blocker 1 — concurrent buffer isolation: addressed.**

Added in-game regression test `/wgen testoverlap <range>` that exercises the real `RentColumnBuffers()` / `ReturnColumnBuffers()` paths under actual worldgen concurrency.

**What the test does:**
1. Resets overlap counters (`stratumColumnsTracked`, `stratumOverlapViolations`)
2. Deletes all chunk columns in the given range
3. Re-generates them via the normal worldgen pipeline (`LoadChunkColumnPriority`), so all concurrent worldgen threads and the full Stratum fast path are exercised exactly as in production
4. Every `RentColumnBuffers()` call adds the buffer to a `ConcurrentDictionary`; every `ReturnColumnBuffers()` removes it. If two live columns ever hold the same buffer, `TryAdd` fails and `stratumOverlapViolations` is incremented
5. After all columns finish loading (+500 ms settle), prints: requested / tracked / violations / verdict

**Runs (Windows x64, 16 cores, dedicated server, same seed and config):**

| Range | Columns | Tracked | Violations | Total time | Result |
|---|---|---|---|---|---|
| 10 | 441 | 441 | **0** | 9 s | PASS |
| 50 | 10 201 | 10 201 | **0** | ~2 min 19 s | PASS |

**Interpretation:**
- `tracked == requested` confirms `GenTerra.OnChunkColumnGen` was invoked for every column (no bypass, no dead queue)
- `violations: 0` across **10 642 total columns** confirms that under sustained concurrent load (6 worldgen threads × ~10 maxThreads scheduler workers = up to ~60 simultaneous live columns in flight) every active column held its own distinct buffer set for the entire call duration
- The 10 201-column run covers a realistic full-region regen (100+ regions worth of work); the zero violation count at this scale rules out the overlap scenario the blocker described

**What remains unchanged:** the compatibility path (`vanilla generate()`) is still not thread-safe by design — it mirrors vanilla's existing behavior and only runs when a Harmony patch requires it. The regression test exercises the fast path, which is the one that matters for the claim "concurrent columns have isolated state".

tehtelev:
Parity test (Blocker 2): parity test with distinct climate and ocean map resolutions

1. A comparison was made between the vanilla path and the `stratumGenerate` fast path. The commands `/wgen parityab 3` and `/wgen parityab 5` were used. The command was entered into the player client, as the results from the console were unstable. Edge smoothing was also disabled in `GenTerra`.
The first run always yielded dozens of differences, as the existing landscape was generated using the standard method. Subsequent reruns yielded:

```
Range: 3 (49 columns generated, 25 compared, smoothing disabled)
Block diffs (fast vs vanilla): 0
Heightmap diffs: 0
Result: PASS (bit-equal)
```

and

```
Range: 5 (121 columns generated, 81 compared, smoothing disabled)
Block diffs (fast vs vanilla): 0
Heightmap diffs: 0
Result: PASS (bit-equal)
```

Here is the test code.


```
private TextCommandResult OnCmdParityAb(TextCommandCallingArgs args)
        {
            int range = (int)args[0];
            if (range < 2)
                return TextCommandResult.Error("Range must be at least 2 to have inner columns for comparison");

            // Stratum: parity tests toggle shared, non-synchronized flags on GenTerra
            // (stratumForceVanillaForParity, stratumParityPreventSmoothing) across an async chain of
            // chunk loads. Running two parity tests concurrently means they stomp on each other's flags
            // mid-flight, producing intermittent, timing-dependent false diffs. Refuse to start a new
            // one while one is still in flight instead of silently corrupting both.
            if (Interlocked.CompareExchange(ref _parityTestRunning, 1, 0) != 0)
            {
                return TextCommandResult.Error("A parity test is already running, wait for it to finish (watch for the report message) before starting another.");
            }

            var gen = api.ModLoader.GetModSystem<GenTerra>();
            if (!gen.stratumUseFastGenerate)
            {
                _parityTestRunning = 0; // Stratum: release the guard on this early-exit path too
                return TextCommandResult.Error("Fast path disabled (generate() patched by a mod) — A/B parity not applicable");
            }

            int chunksize = GlobalConstants.ChunkSize;
            GetCoordsFromRange(range, args.Caller.Player, out List<Vec2i> coords, out _);
            GetCoordsFromRange(range - 1, args.Caller.Player, out List<Vec2i> innerCoords, out _);

            int yChunks = api.WorldManager.MapSizeY / chunksize;
            var blockSnapshots = new Dictionary<Vec2i, int[][]>();
            var rainSnapshots = new Dictionary<Vec2i, ushort[]>();
            var terrainSnapshots = new Dictionary<Vec2i, ushort[]>();
            long blockDiffs = 0;
            long heightDiffs = 0;
            Caller caller = args.Caller;

            bool wasAutogenEnabled = api.WorldManager.AutoGenerateChunks;
            api.WorldManager.AutoGenerateChunks = false;
            GenTerra.stratumParityPreventSmoothing = true;

            void GenerateSequentially(bool forceVanilla, List<Vec2i> columns, int currentIndex, Action done)
            {
                if (currentIndex >= columns.Count)
                {
                    done();
                    return;
                }

                gen.stratumForceVanillaForParity = forceVanilla;
                Vec2i c = columns[currentIndex];
                api.WorldManager.DeleteChunkColumn(c.X, c.Y);

                api.WorldManager.LoadChunkColumnPriority(c.X, c.Y, new ChunkLoadOptions()
                {
                    OnLoaded = () =>
                    {
                        IServerChunk chunk0 = api.WorldManager.GetChunk(c.X, 0, c.Y);

                        if (forceVanilla)
                        {
                            if (innerCoords.Contains(c))
                            {
                                int[][] aBlocks;
                                ushort[] aRain;
                                ushort[] aTerrain;
                                lock (blockSnapshots)
                                {
                                    aBlocks = blockSnapshots[c];
                                    aRain = rainSnapshots[c];
                                    aTerrain = terrainSnapshots[c];
                                }

                                for (int cy = 0; cy < yChunks; cy++)
                                {
                                    var chunk = api.WorldManager.GetChunk(c.X, cy, c.Y);
                                    chunk.Unpack();
                                    var b = chunk.Data;
                                    for (int i = 0; i < b.Length; i++)
                                        if (b[i] != aBlocks[cy][i]) Interlocked.Increment(ref blockDiffs);
                                }

                                var bRain = chunk0.MapChunk.RainHeightMap;
                                var bTerrain = chunk0.MapChunk.WorldGenTerrainHeightMap;
                                for (int i = 0; i < bRain.Length; i++)
                                {
                                    if (bRain[i] != aRain[i]) Interlocked.Increment(ref heightDiffs);
                                    if (bTerrain[i] != aTerrain[i]) Interlocked.Increment(ref heightDiffs);
                                }
                            }
                        }
                        else
                        {
                            if (innerCoords.Contains(c))
                            {
                                var arr = new int[yChunks][];
                                for (int cy = 0; cy < yChunks; cy++)
                                {
                                    var chunk = api.WorldManager.GetChunk(c.X, cy, c.Y);
                                    chunk.Unpack();
                                    arr[cy] = new int[chunk.Data.Length];
                                    for (int i = 0; i < chunk.Data.Length; i++)
                                        arr[cy][i] = chunk.Data[i];
                                }

                                ushort[] rainH = new ushort[chunk0.MapChunk.RainHeightMap.Length];
                                Array.Copy(chunk0.MapChunk.RainHeightMap, rainH, rainH.Length);
                                ushort[] terrainH = new ushort[chunk0.MapChunk.WorldGenTerrainHeightMap.Length];
                                Array.Copy(chunk0.MapChunk.WorldGenTerrainHeightMap, terrainH, terrainH.Length);

                                lock (blockSnapshots)
                                {
                                    blockSnapshots[c] = arr;
                                    rainSnapshots[c] = rainH;
                                    terrainSnapshots[c] = terrainH;
                                }
                            }
                        }

                        GenerateSequentially(forceVanilla, columns, currentIndex + 1, done);
                    }
                });
            }

            GenerateSequentially(false, coords, 0, () =>
            {
                GenerateSequentially(true, coords, 0, () =>
                {
                    gen.stratumForceVanillaForParity = false;
                    api.WorldManager.AutoGenerateChunks = wasAutogenEnabled;
                    GenTerra.stratumParityPreventSmoothing = false;

                    string msg = "[Parity A/B Sequential (inner only)]\nRange: " + range + " (" + coords.Count + " columns generated, " + innerCoords.Count + " compared, smoothing disabled)\n" +
                                 "Block diffs (fast vs vanilla): " + blockDiffs + "\n" +
                                 "Heightmap diffs: " + heightDiffs + "\n" +
                                 "Result: " + (blockDiffs == 0 && heightDiffs == 0 ? "PASS (bit-equal)" :
                                               (blockDiffs <= innerCoords.Count * 8 ? "PASS with FP-boundary tolerance" : "FAIL"));
                    if (caller.Player is IServerPlayer sp)
                        sp.SendMessage(caller.FromChatGroupId, msg, (blockDiffs == 0 && heightDiffs == 0) ? EnumChatType.CommandSuccess : EnumChatType.Notification);
                    else
                        api.Logger.Notification(msg);

                    // Stratum: release the guard only after everything (including the flag resets above)
                    // is fully done, so the next /wgen parityab genuinely cannot start until this one's
                    // shared-flag mutations have all settled.
                    _parityTestRunning = 0;
                });
            });

            return TextCommandResult.Success("Parity A/B sequential (inner only) started on " + coords.Count + " columns. Report will follow.");
        }
```

2. A `stratumVanillaGenerateLock` lock was also added to the vanilla one to ensure that all fields were used by only one thread.

3. The calculation of `b` has been returned to vanilla via Math.Floor, as it always gives the same result for any numbers.

Zaldaryon (changes requested):
Re-reviewed at dd41ab4d. All three blockers from the previous review are genuinely fixed, and I verified each against the actual generated source rather than the diff text or the description of it in the comments.

The buffer publication race is closed on the path that matters. `OnChunkColumnGen`'s smoothing loop now writes into the rented `buf.TaperMap` array and reads a call local `currentBorderIndicesByCardinal` array instead of the shared instance fields, and `stratumGenerate` pulls its buffer from the `[ThreadStatic]` per thread frame stack and shadows every one of `taperMap`, `columnResults`, `layerFullySolid`, and `layerFullyEmpty` with locals bound to that buffer. Two concurrent columns on the fast path no longer touch each other's arrays.

The ocean map bug is fixed, and it turns out to be narrower than the original finding stated: only the Z offset on the two bottom samples used `cfac` instead of `ofac`, in both `generate` and `stratumGenerate`. Both are back to `ofac` now, and `generate`'s ocean block is byte for byte identical to vanilla again. Worth flagging for anyone testing this locally: `TerraGenConfig.cs` sets both `climateMapScale` and `oceanMapScale` to 32 by default, so `ofac == cfac` on a stock world and this bug produced zero visible difference unless a mod changes one of those scales. That is exactly why it needed a parity test with distinct resolutions rather than an eyeball check, and also why it would be easy to reintroduce without anyone noticing.

The nested buffer frame restoration is fixed. The `finally` block now pops, returns the buffer, and republishes the previous frame's buffers and border indices if one existed, using a snapshot taken before the new frame was pushed.

The scheduler oversubscription from the follow up comment is fixed too. `TryExecuteTaskInline` unconditionally returns `false` now, so no caller can run a task body outside the worker array. I checked `ParallelOptions.EffectiveMaxConcurrencyLevel` and confirmed the code comment claiming `MaxDegreeOfParallelism` does not need to be set separately is correct, and traced `TaskReplicator`'s two execution modes to rule out a deadlock from disabling inline execution: only the root replica blocks synchronously, and it blocks the calling worldgen thread, not a worker thread.

I rebuilt Release from a clean checkout at this head. Zero errors, and the only warnings anywhere near this file are the pre-existing NU1904 advisory, unrelated to this change.

Two new findings came out of reading the fix itself, both in the same file.

### Blocking: the pre-lock publish is not covered by the new lock

`OnChunkColumnGen` calls `stratumPublishColumnBuffers(buf)` and sets `borderIndicesByCardinal` at lines 268 and 269, before the `try` block and outside any lock. The vanilla compatibility path only takes `stratumVanillaGenerateLock` later, at line 374, and republishes the same fields again at line 376 right before calling `generate`. That second, locked publish protects a call to `generate` against another locked caller, but nothing stops a second thread's unlocked publish at its own line 268 from landing while the first thread is still inside its locked call to `generate`, which reads `taperMap`, `columnResults`, `layerFullySolid`, and `layerFullyEmpty` directly as instance fields for the whole method (for example lines 498, 523, and 530 in the generated file). Two threads both on the compatibility path, one waiting on the lock while the other runs `generate`, can still stomp each other's buffers between the unlocked publish and the lock being acquired. The fix is to move the publish and the border index assignment at lines 268 and 269 down into the lock, immediately before the call to `generate`, the same way the buffers are already handled at line 376.

This only matters when the compatibility path is active, which requires a mod to have patched `generate`. It is worth being explicit that this can happen more often than the `stratumUseFastGenerate` flag alone suggests, since the code's own comment on that flag documents that mods commonly patch the compiler generated lambda inside `generate` via Harmony rather than the method itself, and the concurrency safety check that is supposed to drop Terrain generation back to single threaded when a mod has patched anything on `GenTerra` walks the type's own methods, not its nested display classes. A mod patching only that lambda can leave Terrain running at full worker concurrency while also taking this exact compatibility path.

### Blocking: disposing the scheduler abandons queued work and hangs the caller

`StratumColumnScheduler`'s worker loop waits on `Monitor.Wait(gate)` until a task is queued or the scheduler is disposed, then returns immediately once disposed is true, before checking whether the queue is actually empty. `initWorldGen` calls `stratumScheduler?.Dispose()` unconditionally every time it runs, and `initWorldGen` is re-entrant: its own comment at line 219 says the landform reset there is useful because `/wgen regen` reloads all the generators. If any task is still sitting in the queue when `Dispose` runs, every worker exits without draining it, and since inline execution is now disabled, the caller waiting on that task through `Task.Wait` or `RunSynchronously` blocks forever. Nothing times out and nothing throws. The fix is to have the worker loop's exit condition check `disposed && queue.Count == 0` instead of just `disposed`, and drain or hand off the queue before the old scheduler's workers exit.

### Non-blocking: the bit-equal parity claim is not literally bit-equal

The `1a81949f` commit correctly restores bit-parity for the Y index by reproducing vanilla's float-add-then-floor instead of the equivalent integer math, and that part is solid, with a good comment explaining why. But the threshold accumulation right after it, at line 982, computes `t[b] * wA[k] + t[b + 1] * wB[k]` with `wA` and `wB` precomputed as `w * (1f - slide)` and `w * slide`. Vanilla computes the mathematically equivalent `weight * GameMath.Lerp(t[b], t[b + 1], slide)`, where `Lerp` is `v0 + (v1 - v0) * t`. Both are correct math, but they are not the same sequence of float operations, so they do not round to the same bits in general. This does not make the PR wrong, and it will rarely flip a threshold comparison, but it does mean the reported zero diff parity runs are evidence that it did not happen to matter on the columns tested, not proof of bit equality.

None of the `/wgen testscheduler`, `/wgen testoverlap`, or `/wgen parityab` commands described in the comments on this PR are in the diff. I checked every commit on this branch and the full repository history for all three names and found nothing. That is exactly the gap that would let someone actually check the two findings above directly instead of by inspection: an overlap run with a mod patching `generate` would exercise the compatibility path the first finding lives in, and a `/wgen regen` mid generation would exercise the second. If those commands get committed as an actual patch, that closes the gap directly instead of asking for trust in a number produced once on one machine.

Marking this changes requested for the two blocking findings above, not the three from the previous review, which are done. The pooling, the frame stack, and the parity fix all show real care.

tehtelev:
Thanks for the incredibly thorough re-review! I've addressed both blocking findings in the latest push:

1. **Pre-lock publish:** Moved `stratumPublishColumnBuffers` and the `borderIndicesByCardinal` assignment strictly inside `stratumVanillaGenerateLock`, immediately before the `generate()` call. I also completely removed the dead unlocked publish on the fast path, since `stratumGenerate` reads directly from the thread-local frame stack.
2. **Scheduler dispose hang:** Updated the `WorkerLoop` exit condition to `disposed && queue.Count == 0`. Workers will now properly drain any remaining queued tasks before exiting when `Dispose()` is called, which prevents the calling thread from hanging indefinitely.

Regarding the non-blocking notes:
* **Bit-equal parity:** Good catch on the float operation sequence (`t[b] * wA + t[b+1] * wB` vs `Lerp`). I'll keep the current optimization as the math is correct and the diff is practically negligible, but I appreciate the precision!

Zaldaryon (changes requested):
Re-reviewed at `9a7e1912`. Same method as before: reconstructed the actual `GenTerra.cs` at this head by applying the patch to the vanilla baseline, rather than reading the patch-of-a-patch.

Blocker 1 from the last round is fixed on the path that mattered. The unconditional unlocked publish before the try block is gone, and the compatibility path now publishes and calls `generate()` inside the same `stratumVanillaGenerateLock` (lines 371-376 of the reconstructed source). Two locked callers can no longer stomp each other's buffers.

Blocker 2 (scheduler dispose hang) is fixed for the case it was reported in, but the fix opens the same hang through an unguarded path.

### Blocking: `QueueTask` has no disposed guard, so a task queued after the drain finishes never runs

`WorkerLoop` now correctly waits for `disposed && queue.Count == 0` before returning, so tasks already queued when `Dispose()` fires do get drained. But once the queue empties and every worker has returned, `QueueTask` still just enqueues into `queue` and calls `Monitor.Pulse(gate)`. There is no live thread left to wake. Since `TryExecuteTaskInline` unconditionally returns `false`, the caller of that task (a worldgen thread waiting via `Task.Wait` inside `Parallel.For`/`TaskReplicator`) blocks forever with no timeout and no exception.

This is reachable, not theoretical. `initWorldGen` disposes and recreates `stratumScheduler` unconditionally on every call, and it is re-entrant through `/wgen regen` (`RegenChunks` in `WgenCommands.cs`). `RegenChunks` waits on `api.Server.PauseThread("chunkdbthread")`, but that only confirms the chunkdbthread's own `paused` flag, not that the fan-out worldgen worker threads actually stopped: `ChunkServerThread.TogglePause` calls `PauseAllWorldgenThreads(1500)` and discards its return value, and `PauseAllWorldgenThreads` has its own 5 second escape hatch that lets a parked worker resume generating even while the pause request is still outstanding. A `/wgen regen` that reloads assets, JSON patches, and landforms easily takes longer than that. So a worldgen thread can still be inside `stratumGenerate`'s `Parallel.For` against the old scheduler when `initWorldGen` tears it down, and the next replica task it queues is silently dropped into a scheduler with zero live workers.

The fix should make `QueueTask` execute the task inline once `disposed` is observed (the caller is already blocked waiting on it, so there is no concurrency budget left to protect at that point), or have `initWorldGen` reuse the existing scheduler when `maxThreads` has not changed instead of disposing and recreating it on every regen.

### Non-blocking: the nested-frame restore still writes shared fields without the lock

`OnChunkColumnGen`'s `finally` block pops the frame and, if there was a previous frame, republishes it via `stratumPublishColumnBuffers` and `borderIndicesByCardinal` unconditionally and outside any lock. If the compatibility path is active and a nested `OnChunkColumnGen` call completes on another thread while the outer call is still inside its locked `generate()`, that unlocked restore can overwrite the buffers `generate()` is reading mid-call. This is the same class of race blocker 1 fixed for the common case, reopened for the nested-call case the frame stack exists to support. Narrow (it needs a mod patching `generate()` plus re-entrant generation), and the fix is one line: take `stratumVanillaGenerateLock` around that restore too, or only restore the fields when the previous frame was itself on the compatibility path.

### Note: the comment justifying the removed publish overstates its case

The comment at the deleted publish site says nothing on the fast path reads those instance fields. `stratumGenerate`'s fallback branch, when no frame owned by `this` is on the stack, still builds its buffer from `this.taperMap`, `this.columnResults`, `this.layerFullySolid`, `this.layerFullyEmpty`. That fallback is unreachable from the registered handler today, so removing the publish does not break Stratum's own path, but the comment overstates it, and `borderIndicesByCardinal` on the fast path now goes stale at whatever `initWorldGen` set it to instead of being refreshed per call, which only matters for a mod reading it through the instance field directly. Worth a corrected comment rather than a functional fix.

Build: swapped the reconstructed head into the bootstrapped tree and ran `dotnet build VSEssentials/VSEssentialsMod.csproj -c Release`. 0 errors, 7 warnings, all pre-existing and unrelated to this file.

Scope: this commit only touches `GenTerra.cs.patch`. The `/wgen testscheduler`, `/wgen testoverlap`, and `/wgen parityab` commands are still not part of the diff, unchanged from the last round.

Marking changes requested for the `QueueTask` gap above. The other two blockers from before are genuinely closed.

tehtelev:
Thanks for catching the edge cases in the scheduler lifecycle and the nested frame restoration! The analysis of the /wgen regen timing and PauseAllWorldgenThreads escape hatch was spot on.

1. **QueueTask disposed guard:** Fixed. `QueueTask` now checks the `disposed` flag under the lock. If the scheduler is already disposed, it falls back to executing the task inline via `TryExecuteTask` outside the lock, preventing the caller from hanging indefinitely. I also took your suggestion to reuse the scheduler in `initWorldGen` if `maxThreads` hasn't changed, which avoids unnecessary thread churn on regen.
2. **Nested-frame restore race:** Fixed. The restoration in the `finally` block is now wrapped in `stratumVanillaGenerateLock`, but only when `!stratumUseFastGenerate`, exactly as you suggested.
3. **Fast-path comment:** Corrected. I updated the comment to acknowledge the fallback branch in `stratumGenerate` and the stale `borderIndicesByCardinal`, clarifying that it only affects unreachable reflection/reverse-patch paths.

Build is clean. Let me know if this covers everything!

tehtelev:
The test was run.

First, `/stratum pregen start radius 100 16000 16000`
Then, in the middle of this, the command was run:
`/wgen regen 5`
and then
`/wgen regen 10`

No errors occurred. I logged in from the client after pregen was complete – everything was fine. No corruption or empty chunks were detected.

`/wgen pregen 10` It works without a hitch. Then I restarted the stratum pregeneration process - everything works as expected.

Zaldaryon (approved):
Re-reviewed at `7a980b22`, "Fix 279 v3". Both blockers from the previous pass are actually fixed, and I don't have a new one to raise. Approving.

Same method as before: reconstructed the real `GenTerra.cs` by applying this commit's patch to the vanilla baseline, not by reading a diff of the `.patch` file against itself. Cross-checked against an independent reconstruction and a from-scratch adversarial pass that rebuilt v2 (`9a7e1912`) and v3 the same way and diffed the two actual sources. Build check: swapped the reconstructed file into a bootstrapped tree and ran `dotnet build VSEssentials/VSEssentialsMod.csproj -c Release`, 0 errors, only the pre-existing unrelated warnings.

**`QueueTask` disposed guard.** Confirmed fixed. `disposed` is now written only under `gate` in `Dispose()` and read only under the same `gate` in `QueueTask` and `WorkerLoop`, so the two can't race. The scenario from the last review (a worker commits to exiting, then a task lands on the normal path and nobody ever runs it) can't happen: a worker can only return while holding `gate` and observing `disposed && queue.Count == 0`, so if `QueueTask` observes `!disposed` under that same lock, at least one worker is guaranteed still live to receive the pulse. And when the scheduler *is* disposed, `QueueTask` now runs the task inline via `TryExecuteTask` before returning, so `Parallel.For`'s wait resolves immediately instead of hanging. `TryExecuteTaskInline` still hard-returns `false` for all normal (non-disposed) execution, so the `maxThreads` cap from the second pass stays enforced; the inline fallback only exists on the teardown path.

Worth noting: the scheduler-reuse change (skip `Dispose()`/recreate when `maxThreads` hasn't changed) means this path is now rarely hit at all in practice, since `maxThreads` is derived from process-lifetime config. `MaximumConcurrencyLevel` is genuinely overridden to `threads.Length`, so the reuse comparison isn't a no-op.

**Nested-frame restore race.** Also fixed. The restore in `OnChunkColumnGen`'s `finally` is now inside `stratumVanillaGenerateLock`, matching the only other writer/reader of those instance fields (`stratumPublishColumnBuffers` + `generate()` at lines 379-384). Gating it on `!stratumUseFastGenerate` is correct rather than a new gap: I traced `stratumGenerate` and confirmed the fast path never touches those instance fields in the first place (it reads its buffer straight off the thread-local frame stack), so there's nothing to restore there, and running the restore on that path would have been the actual bug. The updated comment on the fast-path branch also holds up: `borderIndicesByCardinal` is now write-only in this file (grepped every reference), so "no internal code relies on the stale instance field" checks out.

**Non-blocking, optional follow-up:** the restore at line 396 uses `previousFrame.StratumBuffers` without the same `StratumOwner == this` check `stratumGenerate` applies when reading its own frame. Since the frame stack is `[ThreadStatic]` and shared across `GenTerra` instances, a same-thread nested call from a *different* `GenTerra` instance would restore that instance's buffers into this one's fields. Needs two live `GenTerra` instances on one server to matter, which doesn't happen today, so not blocking, just a one-line hardening (`hadPrevious && previousFrame.StratumOwner == this && !stratumUseFastGenerate`).

**Minor:** `git diff --check` flags a new trailing-whitespace line this commit adds to the patch file itself. Doesn't affect the generated source, just worth cleaning up before merge.

Same caveat as every prior pass: the Linux smoke test is still blocked by the unrelated `Mono.Cecil` resolver issue (#276), so this is a build-level guarantee, not a runtime one. The `/wgen regen` manual test in the last comment is useful supporting evidence but wasn't independently reproduced here; the approval above rests on the source-level correctness argument, not on that test result.

Good work tracking down the actual disposal race instead of just patching the symptom.

Issue #262, opened 2026-08-17
Split entity movement from AI, collision, and pathfinding in the EAR throttle


Evansch0:
### Problem

The EAR throttle slows an entity as one unit. AI, collision, and pathfinding run more often than they need to at distance, player-distance checks run every tick, and entity updates can stack within a tick and cause spikes.

### Proposed change

1. Run movement and physics every tick so they stay smooth, and gate the heavier AI, collision, and pathfinding work behind the distance interval:

   ```
   TickMovement(value, dt);
   if (EarDue(value.EntityId, server.TickPosition, interval))
       TickAIPathfindingAndCollisions(value, accumulatedDt);
   ```

2. Cache the nearest-player distance and reuse it until the player moves past a threshold, and stagger per-entity work across ticks by entity id:

   ```
   double nearestSq = GetNearestPlayerSq(value);
   if ((server.TickPosition + value.EntityId) % GetTickInterval(nearestSq) == 0)
       value.OnGameTick(dt);
   ```

The snippets are illustrative. The real change is larger and the exact scope is not yet clear.

### Alternatives considered

- Keep throttling the whole entity as one unit.
- Only tune the existing distance intervals.

### Scope

- [x] This is a server-side change. It does not require a custom client.

trevorftp:
I’d like to look more into performance on this matter, I know the systems can be improved greatly as it was one of my first implementations with Stratum.

PR #291, opened 2026-09-02
Skip server-side pose matrices until something reads them

Michael-Andrzejewski:
## Summary

`ServerAnimator` no longer recomputes the whole skeleton (`calculateMatrices`) on every tick. It marks the poses stale in `calculateMatrices` and recomputes them on the first read through `GetAttachmentPointPose` or `GetPosebyName`. New config switch `Performance.EntityTicking.LazyServerPoses` (default `true`, hot-reloadable via `/stratum set`) turns it off. `AnimationManager.OnServerTick` also drops a LINQ `Any` that allocated an enumerator and a closure per entity per tick.

Why: every player has `requirePosesOnServer = true`, so vanilla walks all 40 seraph elements, blends every active animation, and multiplies matrices for every player on every server tick, plus for every dead entity. The server only reads those poses in a handful of places (boat seats, rope points, immersive first person eye position, pouring, fishing pole), and only on the tick something happens.

Compatibility:

- Stock client, server-only change.
- Mods that call `GetAttachmentPointPose` or `GetPosebyName` on the server get fresh poses. Checked Overhaullib / CombatOverhaul: its server hit detection goes through `GetAttachmentPointPose`; its `RootPoses` reads are client-only and structural.
- A mod that reads `AnimatorBase.RootPoses` element matrices directly on the server without calling a getter first would see stale poses. `LazyServerPoses: false` restores vanilla behaviour.
- Readers on a physics thread take a lock for the recompute; the vanilla scratch matrices are already thread static.

## Type

- [ ] Bug fix
- [x] Performance
- [ ] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

## Performance numbers

Micro-benchmark on the real `seraph.json` against `VintagestoryAPI.dll` 1.22.7 (desktop CPU, .NET 10, 30000 frames per case after warm up):

| active animations | with matrices | without | 650 players, per tick |
| --- | --- | --- | --- |
| none | 7.24 us | 0.01 us | 4.7 ms |
| idle1 | 5.51 us | 0.05 us | 3.6 ms |
| walk + idle1 | 6.96 us | 0.08 us | 4.5 ms |
| walk + idle1 + held item | 7.41 us | 0.14 us | 4.8 ms |

The 650-bot timings report from two months ago had `entity.tick.players` at 13.1 ms per tick average. This is 30 to 40 percent of that, on a machine that is probably faster than the server.

Equivalence check: eager and lazy animators ticked with the same random animation set for 6000 frames, attachment points read on random frames: 1842 reads, maximum absolute matrix difference 0.0, 6000 computes eager vs 614 lazy.

Real server (this branch, launcher with the built assemblies overlaid), one player, same spot, standing still, same machine state. Absolute numbers are high because the desktop had other work running; the comparison is A against B:

| run | LazyServerPoses | ticks | entity.type.player avg | max |
| --- | --- | --- | --- | --- |
| A | true | 5457 | 0.233 ms | 3.8 ms |
| B | false | 12694 | 0.290 ms | 8.4 ms |

57 us less per player per tick with lazy poses, about 20 percent of the player entity tick on that machine. Run B was otherwise the lighter run (creature tick 0.36 vs 0.41 ms, physics 5.9 vs 6.5 ms), so the difference is not the world getting busier. The bot swarm is still the number that matters for the 650-player case: compare `entity.type.player` in `/stratum timings report` with `Performance.EntityTicking.LazyServerPoses` true vs false.

Functional check on the same session: creative flight loading chunks, spawned and killed mobs, corpse harvesting, normal play, both settings; no exceptions in the server log.

## Related issues

Related to #262 (entity tick cost), not a fix for it.

Side note: `scripts/smoke-test.ps1` prefers `StratumServer/bin/Release/net10.0/VintagestoryServer.exe` when it exists. After the launcher's first run that file is the vanilla server it downloaded, so from then on the smoke test boots vanilla, not Stratum, and on a dev build (`EmbedPatchedFiles=false`) the launcher overlays nothing either. Worth pointing the smoke test at `StratumServer.exe` after a `StratumDeploy` overlay, or checking for the Stratum banner in the log.

### DIFF
--- a/patches/VintagestoryApi/Common/Model/Animation/AnimationManager.cs.patch
+++ b/patches/VintagestoryApi/Common/Model/Animation/AnimationManager.cs.patch
@@ -1,5 +1,5 @@
 diff --git a/VintagestoryApi/Common/Model/Animation/AnimationManager.cs b/VintagestoryApi/Common/Model/Animation/AnimationManager.cs
-index 573ec12..de0f300 100644
+index 573ec12..74acb9a 100644
 --- a/VintagestoryApi/Common/Model/Animation/AnimationManager.cs
 +++ b/VintagestoryApi/Common/Model/Animation/AnimationManager.cs
 @@ -27,11 +27,23 @@ namespace Vintagestory.API.Common
@@ -314,7 +314,7 @@ index 573ec12..de0f300 100644
                      val.Value.StartFrameOnce = anim.CurrentFrame;
                  }
  
-@@ -446,14 +595,16 @@ namespace Vintagestory.API.Common
+@@ -446,15 +595,28 @@ namespace Vintagestory.API.Common
          /// The event fired at each server tick.
          /// </summary>
          /// <param name="dt"></param>
@@ -327,14 +327,27 @@ index 573ec12..de0f300 100644
              {
 -                Animator.CalculateMatrices = !entity.Alive || entity.requirePosesOnServer;
 -                Animator.OnFrame(ActiveAnimationsByAnimCode, dt);
+-                AdjustCollisionBoxToAnimation = ActiveAnimationsByAnimCode.Any(anim => anim.Value.AdjustCollisionBox);
 +                stratumAnimator.CalculateMatrices = !entity.Alive || entity.requirePosesOnServer;
 +                stratumAnimator.OnFrame(ActiveAnimationsByAnimCode, dt);
-                 AdjustCollisionBoxToAnimation = ActiveAnimationsByAnimCode.Any(anim => anim.Value.AdjustCollisionBox);
++                // Stratum start: plain loop, the LINQ Any allocated an enumerator and a closure per entity per tick
++                bool stratumAdjustCollisionBox = false;
++                foreach (KeyValuePair<string, AnimationMetaData> stratumAnim in ActiveAnimationsByAnimCode)
++                {
++                    if (stratumAnim.Value.AdjustCollisionBox)
++                    {
++                        stratumAdjustCollisionBox = true;
++                        break;
++                    }
++                }
++                AdjustCollisionBoxToAnimation = stratumAdjustCollisionBox;
++                // Stratum end
              }
  
              runTriggers();
          }
-@@ -495,10 +646,11 @@ namespace Vintagestory.API.Common
+ 
+@@ -495,10 +657,11 @@ namespace Vintagestory.API.Common
              }
          }
  
@@ -346,7 +359,7 @@ index 573ec12..de0f300 100644
          }
  
          private void runTriggers()
-@@ -537,11 +689,11 @@ namespace Vintagestory.API.Common
+@@ -537,11 +700,11 @@ namespace Vintagestory.API.Common
              }
          }
  
--- a/patches/VintagestoryApi/Common/Model/Animation/AnimationManager.cs.patch
+++ b/patches/VintagestoryApi/Common/Model/Animation/AnimationManager.cs.patch
@@ -1,5 +1,5 @@
 diff --git a/VintagestoryApi/Common/Model/Animation/AnimationManager.cs b/VintagestoryApi/Common/Model/Animation/AnimationManager.cs
-index 573ec12..74acb9a 100644
+index 573ec12..21aa3dc 100644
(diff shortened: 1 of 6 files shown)


patches/VintagestoryApi/Common/Model/Animation/AnimationManager.cs.patch
```diff
@@ -327,14 +327,27 @@ index 573ec12..de0f300 100644
              {
 -                Animator.CalculateMatrices = !entity.Alive || entity.requirePosesOnServer;
 -                Animator.OnFrame(ActiveAnimationsByAnimCode, dt);
+-                AdjustCollisionBoxToAnimation = ActiveAnimationsByAnimCode.Any(anim => anim.Value.AdjustCollisionBox);
 +                stratumAnimator.CalculateMatrices = !entity.Alive || entity.requirePosesOnServer;
 +                stratumAnimator.OnFrame(ActiveAnimationsByAnimCode, dt);
-                 AdjustCollisionBoxToAnimation = ActiveAnimationsByAnimCode.Any(anim => anim.Value.AdjustCollisionBox);
++                // Stratum start: plain loop, the LINQ Any allocated an enumerator and a closure per entity per tick
```
trevorftp: The LINQ path can box the enumerator but this lambda doesn't capture anything so there isn't a closure per tick. Loop is fine, just fix the comment.

patches/VintagestoryApi/Common/Model/Animation/ServerAnimator.cs.patch
```diff
@@ -0,0 +1,54 @@
...
++        }
++
++        protected override void StratumEnsurePosesFresh()
++        {
++            if (!stratumPosesStale) return;
++
++            // Readers can come from a physics thread while the main thread ticks the animation.
++            // Two readers must not recompute at the same time; the scratch matrices are thread
++            // static but jointsDone and the poses are shared.
++            lock (stratumPoseLock)
```
trevorftp, line 54: This lock only covers readers? OnFrame is still mutating the animation state while a physics thread can come in here and recompute from it I would think we need the writer synced too.

patches/VintagestoryApi/Common/Model/Animation/ServerAnimator.cs.patch
```diff
@@ -0,0 +1,54 @@
...
++        protected override void calculateMatrices(float dt)
++        {
++            if (StratumLazyPoses && CalculateMatrices)
++            {
++                stratumPosesStale = true;
++                stratumLastDt = dt;
++                return;
++            }
++
++            base.calculateMatrices(dt);
```
trevorftp, line 44: Should clear stratumPosesStale here too. If lazy poses gets turned off while this is stale, the next getter recalculates it again anyway

trevorftp (changes requested):
Look's reasonable to me, the idea makes sense and numbers look promising.

patches/VintagestoryApi/Common/Model/Animation/AnimationManager.cs.patch
Michael-Andrzejewski: Fixed the comment!

patches/VintagestoryApi/Common/Model/Animation/ServerAnimator.cs.patch
Michael-Andrzejewski: Fixed, OnFrame takes the lock now and avoids the overlap

patches/VintagestoryApi/Common/Model/Animation/ServerAnimator.cs.patch
Michael-Andrzejewski: Fixed, eager path now clears the flag here too

Michael-Andrzejewski:
Fixes made and ready for re-review

trevorftp:
Almost, but the first stratumPosesStale check is still outside the lock. Make that field volatile or move the check inside, otherwise the reader can still miss the writer setting it.

Michael-Andrzejewski:
Thanks, I have now made the field volatile to address this. Ready for re-review again

PR #298, opened 2026-09-07
Stack a role's name prefixes, ordered by Priority

### LINKED ISSUE #274: Stack multiple role prefixes on one player, ordered by Priority
### Problem

A role supports only one prefix. The `Priority` property exists but is ignored, and the server appears to use the last matching entry in the config. An admin cannot mark a player as a moderator without overwriting the prefix they carry for gameplay reasons.

### Proposed change

Allow multiple prefixes on the same role and render them stacked before the name, using the existing `Priority` value to order them, for example `[Moderator][XYZ Villager] Name`. Honoring `Priority` for ordering is the small first step, since the field is already in the schema.

### Alternatives considered

Create a separate combined role per player. That does not scale and forces a choice between the moderator tag and the gameplay tag. A related option is to allow multiple roles per player, which would be its own change.

### Scope

Zaldaryon:
## Summary

A role can now carry more than one name prefix. The tags stack in front of the name, ordered by the `Priority` field that was already in the schema, so an operator can put a `[Moderator]` badge in front of a player without losing the tag their gameplay role already gives them.

### Problem

`Appearance.RolePrefixes.Roles` maps a role code to one prefix. `Priority` exists on each prefix but has never changed anything, and an admin who wants a player to show both a staff tag and a gameplay tag has to build a combined role for every pairing, which does not scale.

### Current behaviour

Two prefix entries written under the same role key do not both survive. `Roles` is a JSON object, Newtonsoft's default duplicate-key handling is "keep the last", so the earlier entry is discarded before Stratum's code runs. And because a `Dictionary` holds at most one value per key, the `OrderByDescending(Priority)` in both the chat formatter and the nametag builder has only ever sorted a one-element sequence. `Priority` has been dead.

### Design

Each role's value is now read and written as a bare object for a single prefix and as an array for several:

```json
"Roles": {
  "admin": { "Tag": "Admin", "Priority": 100 },
  "villagemod": [
    { "Tag": "Moderator", "Color": "#4cc9f0", "Priority": 100 },
    { "Tag": "XYZ Villager", "Color": "#9bd77e", "Priority": 10 }
  ]
}
```

Scalar-or-array under one key is the common convention for exactly this shape (npm `bin`, GitHub Actions `on`, ESLint `extends`): the syntax an operator reaches for first is the syntax that works, and a config that only ever wanted one tag never has to learn the array form.

The write side is asymmetric on purpose. `StratumRuntime.LoadOrCreateConfig` rewrites `stratum.json` on every boot, so if a one-prefix role serialised back as a one-element array, every existing server's config would be reshaped on the next start. `StratumRolePrefixListConverter` writes a single prefix back as a bare object, so an untouched role stays byte-for-byte as it was. Only a role you actually give a second tag becomes an array. This also means no `ConfigVersion` bump and no migration: old files load unchanged.

### Change

`sources/VintagestoryLib/Vintagestory.Server/`:

- **`StratumAppearanceConfig.cs`**: `Roles` is now `Dictionary<string, StratumRolePrefixList>`. `StratumRolePrefixList` is a thin holder with a class-level `[JsonConverter]` that reads object-or-array and writes the shape back. `StratumRolePrefixesConfig.ResolveFor(roleCode)` returns the enabled prefixes for a role, highest `Priority` first (a stable sort, so equal priorities keep config-file order). The role-code match is a hand-written case-insensitive scan because `ObjectCreationHandling.Replace` rebuilds the dictionary on load and drops its `OrdinalIgnoreCase` comparer.
- **`StratumChatFormatter.cs`**: `TryFormat` renders each tag through the existing `Format` / colour / weight path and concatenates them, one space before the name. `FindPrefix` is deleted.
- **`StratumNametags.cs`**: `ApplyNametagPrefix` concatenates each tag through `PrefixFormat` into the one nametag string. `FindRolePrefix` is deleted.
- **`CmdStratum.cs`**: `/stratum chat` now prints each role's rendered tag stack plus every tag's priority and colour, disabled ones included.

A role with one prefix renders identically to before on all three surfaces. A role with none, or with every tag disabled, renders a bare name, same as today.

`FindPrefix` and `FindRolePrefix` were the same resolution logic in two files; they are replaced by the single `ResolveFor`.

### Regression proof

`scripts/smoke-test.sh` seeds `stratum.json` before boot with a `suvisitor` role carrying two prefixes as an array, deliberately in the wrong file order (`XYZ Villager` priority 10 before `Moderator` priority 100), and an `admin` role left as a bare object. The seeded file makes the boot config load run the converter through the real `JObject.ToObject<StratumConfig>` path. After `RunGame` the probe asserts, on a real server boot:

- `/stratum chat` renders `[Moderator][XYZ Villager]` for `suvisitor`: the stack is in `Priority` order, not file order.
- the boot rewrite of `stratum.json` keeps `suvisitor` an array and `admin` a bare object.

Verified on `984c3d6`: `bash scripts/smoke-test.sh` reaches RunGame and passes all 9 console probes, and the on-disk `stratum.json` after the run has `"suvisitor": [` and `"admin": {`.

The chat formatter and the nametag string against a live joined player are covered by a scenario in the private Atlas harness (`RolePrefixScenarios`). That scenario is written and compiles, but the harness's synthetic-client bridge is still pinned to Vintage Story 1.22.6 and cannot join a 1.22.7 server, so it does not run yet; the same gap blocks every existing scenario in that harness. The console probe covers the ordering and stacking end to end in the meantime.

### Gates

- `dotnet build VintageStory.slnx -c Release` (two-pass, `EmbedPatchedFiles=true`): green, zero new warnings on the four touched files.
- `bash scripts/smoke-test.sh`: PASS, server reaches RunGame, all 9 console probes verified.
- `scripts/extract-patches.sh`: no `patches/` change. The whole diff is `sources/` and docs.

### Limitations

- In the nametag, all tags share one colour. It comes from `Nametags.EntitlementColorByRole`, mapped from the player's role to a Vintage Story entitlement code, because the stock client has no way to colour part of a nametag. Chat colours each tag on its own. This is a protocol limit, not a shortcut, and it is documented.
- A long stack crowds the chat line and widens the in-world tag. The docs say keep it to two or three; nothing enforces a cap.
- Giving one player more than one role is a separate, larger change and stays out of scope. `Fixes #274` closes the stacking request; the multi-role idea is noted in the issue for its own follow-up.

### Observations, offered as follow-ups rather than bundled here

- `/stratum chat`'s `Prefix format` row prints the literal `{tag}` token, and the server console's result logger runs `string.Format` on every message unconditionally, so that one row logs a `Couldn't write to log file` error. Reproducible on `indev` today by typing `/stratum chat` at the console. The fix belongs in the vanilla console path, not here; the smoke probe reads the row on its own line and is not affected.
- `ObjectCreationHandling.Replace` dropping the dictionary comparer on load also affects `StratumNametagsConfig.EntitlementColorByRole`, which still does a case-sensitive `TryGetValue`. A role code cased differently from its config key gets no entitlement colour. Pre-existing, separate.

## Type

- [ ] Bug fix
- [ ] Performance
- [x] New feature
- [x] Refactor or cleanup
- [ ] Docs or build

## Checklist

- [x] `scripts/extract-patches.sh` ran clean.
- [x] Every vanilla edit has a `// Stratum` marker. (No vanilla edits: the four changed files are all Stratum-only `sources/`.)

## Related issues

Fixes #274

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,8 @@ Anything else is forwarded to the server, such as `--port` or `--dataPath`.
 ## Commands
 
 Stratum adds a set of in-game commands on top of vanilla. See
-[docs/commands/kits.md](docs/commands/kits.md) for `/kit` and `/kitedit`.
+[docs/commands/kits.md](docs/commands/kits.md) for `/kit` and `/kitedit`, and
+[docs/role-prefixes.md](docs/role-prefixes.md) for stacking role name prefixes.
 
 ## Build
 
--- a/docs/role-prefixes.md
+++ b/docs/role-prefixes.md
@@ -1,0 +1,116 @@
+# Role name prefixes
+
+Stratum can put a coloured tag in front of a player's name in chat and above
+their character, chosen by the player's role. A role can carry more than one
+tag, and they stack in a fixed order. Implementation:
+[`StratumChatFormatter.cs`](../sources/VintagestoryLib/Vintagestory.Server/StratumChatFormatter.cs)
+for chat, [`StratumNametags.cs`](../sources/VintagestoryLib/Vintagestory.Server/StratumNametags.cs)
+for the in-world nametag, both reading
+[`StratumAppearanceConfig.cs`](../sources/VintagestoryLib/Vintagestory.Server/StratumAppearanceConfig.cs).
+
+Server side only. No custom client: the tags ride in on the normal chat and
+nametag protocol, so every stock client renders them.
+
+## Where the config lives
+
+`<dataPath>/stratum.json`, under `Appearance.RolePrefixes`:
+
+```json
+"RolePrefixes": {
+  "Enabled": true,
+  "Format": "[{tag}]",
+  "Roles": {
+    "admin": { "Tag": "Admin", "Color": "#ff5f57", "Bold": true, "Priority": 100 },
+    "villagemod": [
+      { "Tag": "Moderator", "Color": "#4cc9f0", "Bold": true, "Priority": 100 },
+      { "Tag": "XYZ Villager", "Color": "#9bd77e", "Bold": false, "Priority": 10 }
+    ]
+  }
+}
+```
+
+| Key | Meaning |
+| --- | --- |
+| `Enabled` | Master switch for chat prefixes. `false` turns every role's chat tag off. The nametag has its own switch, `Nametags.Enabled` / `Nametags.ApplyRolePrefix`. |
+| `Format` | Wraps each tag. `{tag}` is the placeholder. `"[{tag}]"` gives `[Admin]`, `"{tag} "` gives `Admin `. |
+| `Roles` | Role code to prefix, or role code to a list of prefixes. |
+
+Each prefix entry:
+
+| Field | Default | Meaning |
+| --- | --- | --- |
+| `Tag` | `Staff` | The text inside `Format`. |
+| `Color` | `#ffffff` | `#rrggbb`, or a plain colour word. Chat only (see below). |
+| `Bold` | `true` | Chat only. |
+| `Priority` | `0` | Higher renders closer to the name's left. |
+| `Enabled` | `true` | `false` drops just this one tag. |
+
+## One tag or several
+
+A role's value is written as a bare object for a single tag and as an array for
+several. Both forms are always accepted on read. On save Stratum writes a
+one-tag role back as a bare object, so a config that predates this feature is
+left exactly as it was. Only a role you actually give a second tag turns into an
+array.
+
+To add a second tag to a role, change its `{ ... }` to `[ { ... }, { ... } ]`.
+
+## Order
+
+Tags render highest `Priority` first, left to right, then the name. With the
+example above, a player in `villagemod` shows as `[Moderator][XYZ Villager]
+Name`. Two tags at the same `Priority` keep the order they appear in the config
+file.
+
+## Spacing
+
+The gap between the tags and the name comes from `Format`. `"[{tag}]"` gives
+`[A][B] Name`; `"[{tag}] "` gives `[A] [B] Name`. There is no separate spacing
+key. The nametag has its own `Nametags.PrefixFormat`, which defaults to
+`"[{tag}] "`.
+
+## Colour
+
+Chat colours each tag on its own, from that tag's `Color` and `Bold`.
+
+The nametag is one colour for the whole string. That colour is not per tag: it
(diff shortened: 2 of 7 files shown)


Pixnop (commented):
The object-or-array read with an object-only write-back is the right shape: no ConfigVersion bump, no migration, and an untouched stratum.json comes back unchanged through the boot rewrite. Folding FindPrefix and FindRolePrefix into one ResolveFor removes a real divergence risk, and the comment at StratumAppearanceConfig.cs:135 on why the linear scan exists will save the next person from turning it back into a broken TryGetValue.

One thing I would change before merge, and it is in the script rather than the C#. scripts/smoke-test.sh:80 writes the seed unconditionally, while line 66 sets own_data=0 when SMOKE_TEST_DATA is supplied and cleanup at line 102 deliberately leaves that directory alone. So SMOKE_TEST_DATA=~/vsserver bash scripts/smoke-test.sh now replaces that server's real stratum.json with the fifteen-line seed, and the boot rewrite then persists defaults for every other section. Until now the script only ever put console.fifo and smoke-test.log in a caller-supplied data path. Gating the seed on own_data, or saving the file aside and restoring it in cleanup, covers it.

Smaller: CmdStratum.cs:465 sorts with its own OrderByDescending instead of going through ResolveFor, so the new probe asserts on the command's copy of the ordering, not on the one chat and nametags use. Is there a shape where /stratum chat can reuse ResolveFor and still list disabled tags?

Pixnop (changes requested):
Switching this to a change request so the state matches the review above: the seed at scripts/smoke-test.sh:80 needs the own_data gate, or a save and restore in cleanup, before this goes in. Everything else stands as comments.

Zaldaryon:
## Re-review requested

Zaldaryon:
## Re-review update

The caller-supplied data path now skips both the role-prefix seed and its private-file assertions. Verified with the existing data directory path: the server reached RunGame and all 8 applicable console probes passed.

Latest fix: `5f4aab4`.

Pixnop (approved):
The seed is gated now. scripts/smoke-test.sh:78 wraps the heredoc and the /stratum chat probe in own_data == 1, the two rewrite assertions moved under the same gate at 234, and the note at 77 records the rule, so a caller-supplied SMOKE_TEST_DATA keeps its own stratum.json and the run stays green on a healthy server. That was the blocker, and it is gone.

CmdStratum.cs:465 still sorts on its own rather than through ResolveFor. Fine to leave as is if folding disabled tags into the resolver is not worth the shape.

Approving.

Issue #277, opened 2026-08-22
Config option to disable friendly fire within a group at runtime


PHoenixOPHury:
### Problem

Players fighting as a team hit and kill each other through friendly fire. In team-versus-team play that is unwanted, and it makes coordinated PvP harder.

### Proposed change

- A runtime config option that stops players in the same group from damaging each other.
- Admin commands, usable at runtime, to add a player to a group without an invite and to stop a player from leaving a group.
- More group configuration options generally.

### Scope

- [x] This is a server-side change. It does not require a custom client.

Zaldaryon:
#300 delivers the first bullet: a runtime `/friendlyfire on|off|toggle|status` toggle, gated behind `stratum.friendlyfire`, that stops players in the same player-created group from damaging each other. It is on by default so nothing changes until an admin turns it off. The block sits in `EntityPlayer.ShouldReceiveDamage`, so it also stops knockback and death, heals still pass, and self damage still works.

The other two asks are a different feature. Forcing a player into a group without an invite and preventing a player from leaving a group are moderation controls over group membership, not combat behaviour, and they need their own design (who can run them, how a forced membership is displayed to the player, whether a locked member can be kicked). I would rather split those into a separate issue than bundle them here. Happy to open it, or leave that to you.

Zaldaryon:
Follow-up to #300, four stacked PRs that close the limitations #300's own description listed. Each is `Part of #277`.

- **#301 (1/4): docs and the shared group check.** Adds `docs/commands/friendlyfire.md`. Routes `StratumPlayerPrivacy`'s map-pin group exception through the shared `StratumPlayerGroups.SharesGroup`; the old local copy counted a `None`-level membership as still in the group, which leaked exact coordinates under group map visibility.
- **#302 (2/4): explosions.** `CreateExplosion` already gets the igniter's uid (bombs record it and pass it through chain reactions); it just never reached the damage path. A group mate's blast now skips the group while the toggle is off, and the igniter still takes their own blast.
- **#303 (3/4): block at the source seams.** Moves the check up from `ShouldReceiveDamage` to the melee interaction handler and the projectile impact path, so a blocked swing costs no weapon durability, plays no slap sound, and no longer combat-tags both players (that last one was a bug: blocked melee tagged, blocked arrows did not). Adds a throttled "that player is in your group" notice to the attacker and a blocked-hit counter in `/friendlyfire status`.
[...]

PR #300, opened 2026-09-07
Group friendly fire: /friendlyfire toggle (part of #277)

Zaldaryon:
## Summary

Adds a runtime toggle that stops players in the same player group from damaging each other. This is bullet one of #277. The admin group commands (force a player into a group, stop a player leaving) are a separate moderation feature and belong in their own issue, see the issue comment.

Friendly fire is **on by default** (`AllowGroupDamage = true`). That matches vanilla and matches how every other server platform reads the phrase: friendly fire on means teammates can hurt each other, turning it off protects the group. Minecraft, Spigot and the rest default it on.

### Change

- **`sources/VintagestoryApi/Server/StratumFriendlyFireHook.cs`** (new): a one-field static bridge, `BlockGroupDamage`. Lives in VintagestoryApi so both VintagestoryLib (the patched `EntityPlayer`) and any future VSEssentials reader can see it, same pattern as `StratumCoopCombatHook` from PR #243.
- **`sources/VintagestoryApi/Server/StratumPlayerGroups.cs`** (new): `SharesGroup`. Two players share a team when they both hold a live membership (`Level != None`) in the same player-created group (`GroupUid > 0`). Default chat groups (uids `<= 0`) are excluded. There is an `IPlayer` overload and a `Dictionary<int, PlayerGroupMembership>` overload so the semantics are testable without a live server. It iterates the smaller membership dictionary.
- **`patches/VintagestoryApi/Common/Entity/EntityPlayer.cs.patch`**: the block sits in `ShouldReceiveDamage`, before the vanilla `base` call and after the existing vanilla PvP gate. Returning `false` there also stops knockback, `DidAttack` bookkeeping, damage-over-time application and death, not just the health change. Guards: server side only (`World?.Api is ICoreServerAPI`, needed for a listen server), `damageSource.Type != Heal` so a group mate can still heal you, and `stratumAttacker != this` so self damage always passes. It uses `damageSource.GetCauseEntity()`, not `CauseEntity`: `CauseEntity` is set only for projectiles, a melee hit leaves it null and puts the attacker in `SourceEntity`, and `GetCauseEntity()` returns `CauseEntity ?? SourceEntity`.
- **`sources/VintagestoryLib/Vintagestory.Server/StratumFriendlyFireConfig.cs`** (new): one property, `AllowGroupDamage`, default `true`.
- **`sources/VintagestoryLib/Vintagestory.Server/StratumFriendlyFireSystem.cs`** (new): owns the config value, pushes it into the hook at startup, on `/friendlyfire`, and after a live `/stratum reload`. Registers `/friendlyfire` with an `OptionalWordRange` of `on|off|toggle|status`, gated behind the `stratum.friendlyfire` privilege, with the standard Stratum command-access and cooldown checks. Mirrors `StratumCoopCombatSystem`.
- **`sources/VintagestoryLib/Vintagestory.Server/StratumConfig.cs`, `StratumCommandAccessCatalog.cs`, `CmdStratum.cs`**: wire the new config block, the `/friendlyfire` access entry, and the reload re-apply next to the existing `CoopCombat` ones.
- **`patches/VintagestoryLib/Vintagestory.Server/ServerSystemCommands.cs.patch`**: `new StratumFriendlyFireSystem(server);` next to the other Stratum systems.
- **`scripts/smoke-test.sh`**: five index-matched probe pairs exercising status, toggle, on and off.

### Gates

- `dotnet build VintageStory.slnx -c Release` then `... -p:EmbedPatchedFiles=true`: green. Zero new warnings on the touched files. The 4 solution warnings are the pre-existing vanilla `NU1904` System.Drawing.Common advisories.
- Standalone group-semantics check against `SharesGroup(Dictionary, Dictionary)`, 9 cases, all pass: shared group, partial overlap, disjoint, default chat group excluded, negative uid excluded, dead `None` membership, empty dict, null dict, op-level membership.
- `bash scripts/smoke-test.sh`: `PASS: server reached RunGame, no fatal errors, 13 console command(s) verified.` All five `/friendlyfire` probe outputs present in the boot log:
  - `/friendlyfire` with the default config: `Group friendly fire is on, players in the same group can damage each other.`
  - `/friendlyfire off`: `Group friendly fire disabled, players in the same group can no longer damage each other.`
  - `/friendlyfire status`: `Group friendly fire is off, players in the same group cannot damage each other.`
  - `/friendlyfire toggle` (from off): `Group friendly fire enabled, players in the same group can damage each other again.`
- `scripts/extract-patches.sh`: only `EntityPlayer.cs.patch` and `ServerSystemCommands.cs.patch` change.

### Follow-ups

Four stacked PRs close the gaps this PR's first version listed as limitations. All `Part of #277`.

- **#301** (docs + shared group check). Adds `docs/commands/friendlyfire.md`. Routes `StratumPlayerPrivacy`'s map-pin group exception through `StratumPlayerGroups.SharesGroup`; the private copy there (`StratumPlayerPrivacy.cs:98`) counted a `None`-level membership as still in the group, leaking exact coordinates under group map visibility.
- **#302** (explosions). An explosion `DamageSource` carries no cause entity, but `CreateExplosion` already receives `ignitedByPlayerUid` (bombs record it and pass it through chain reactions). #302 resolves the igniter there and drops the igniter's group mates from the hurt list; the igniter still takes their own blast.
- **#303** (block at the source). The hook here in `ShouldReceiveDamage` runs after the slap sound, the weapon durability tick (`Collectible.OnAttackingWith`), and the interact event that combat-tags both players. #303 moves the block up to the melee interaction handler and the projectile impact path, adds a per-attacker throttled "that player is in your group" notice, and stops the combat tag on a blocked hit (blocked melee wrongly tagged before; blocked arrows did not). This hook stays as the catch-all for mod-authored damage.
- **#304** (Harmony conflict warning). When the toggle is on, logs a warning naming any mod that Harmony-patches a method the toggle relies on. A mod that replaces one of those and skips the original still bypasses this, same as it bypasses vanilla PvP gating; nothing can hard-prevent that, so the goal is to make it visible.

Damage over time: the check runs once, when the hit that starts the effect lands, which matches Minecraft. Vanilla has no player-inflicted damage over time, so this only matters with a mod that adds one; #301 states the rule in the config comment. Full attribution into `DamageOverTimeEffect`, whose `ToBytes` is a client-visible wire format, would be its own issue.

### Testing

Private Atlas suite `research/atlas-tests/stratum-pr-validation/FriendlyFireScenarios.cs` (new): 4 of 7 scenarios pass on a real 1.22.7 build via `dotnet test`: the `/friendlyfire` command moves the cross-assembly hook, the command is blocked when its access config is off, `/stratum reload` resyncs a drifted hook, and the `SharesGroup` dictionary rule rejects dead, partial, disjoint, and default-chat-group memberships. The three scenarios that need a second connected player (melee lands or not, heal, explosion) fail with `AtlasSetupException: ... did not finish joining the world`, the same 1.22.7 synthetic-client join gap that blocks the #274 suite, not this change.

## Type

- [ ] Bug fix
- [ ] Performance
- [x] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

- [x] `scripts/extract-patches.sh` ran clean (only the two patches this change touches).

## Related issues

Part of #277


Pixnop (commented):
The wiring is complete and matches StratumCoopCombatSystem line for line, which is what I want from a toggle like this. Good call on GetCauseEntity() over CauseEntity, and on the GroupUid > 0 plus Level != None rule in StratumPlayerGroups.cs:31. I also checked the boot order: ServerSystemStratum sits before ServerSystemCommands in the Systems array, so stratum.json is loaded before the system reads it.

One thing I would like settled before this lands on its own. The block sits in ShouldReceiveDamage, so a blocked melee swing has already ticked weapon durability through OnAttackingWith and already raised OnPlayerInteractEntity. On a server with CombatLog.Enabled, StratumCombatLogSystem.cs:86 tags both players for a hit that dealt zero damage, and Penalty defaults to Kill, so a group mate can tag you and a disconnect kills you. #303 moves the block up and fixes exactly that. Is #300 meant to be mergeable alone, or does the stack land together?

A design question. SharesGroup counts any shared group as a team, including one whose JoinPolicy is everyone. On a PvP server that runs a public channel as a player group, /friendlyfire off hands mutual invulnerability to anyone who joined it. Intended reading of "same group", or should open groups be excluded like the default chat groups are?

On the join gap in your Atlas runs: it is not 1.22.7, it is our own first-packet gate. Since c7c83f6 (Aug 24) ServerMain drops any connection whose first data packet is not a ServerQuery or a LoginTokenQuery, silently, before a ConnectedClient exists. The vanilla client always sends the token query first; the synthetic client sends Identification straight away. I hit the same nine failures in StratumParity on stable v1.22.7-stratum.2, and the same suite passes 17/17 on vanilla 1.22.7. The fix is on the Atlas side (send the token query first), so your two-player scenarios should unblock without touching Stratum.

Pixnop:
Read the five as a set. It is a tree, not a pile: #300 is the root, #301, #302 and #303 chain off it, and #304 branches straight from #300. Real dependencies are looser than the declared bases (#302 needs only #300, #303 needs #301 only because it edits the page that PR creates), so merge in the declared order and in one window rather than reordering, which would conflict on the StratumFriendlyFireConfig header comment and the docs page.

Defaults are right: AllowGroupDamage stays true, every seam starts with the static bool, and stratum.friendlyfire is granted to no role, so an existing server sees no change. The one exception is the map-pin fix in #301, which is not gated by the toggle and deserves its own changelog line. Vanilla clients are unaffected: no new packets, no wire change.

Three fixes before the set merges: the #301 page describes explosions, multi-point enforcement and the Harmony warning as shipped; the on/off wording of that warning is inverted in the page and in the #304 comment; and the threading in #303. Plus the uncovered vectors named on #301, which want a line in the page, not code.

Zaldaryon:
## Re-review status

The root toggle is intentionally held with the #277 stack. The blocked melee path is corrected in #303, and Harmony warning refresh is corrected in #304. Please re-review the complete stack before merging #300 into `indev`.

Pixnop (approved):
Approving the root on the terms you set: it lands with the stack, not alone. #303 is approved and removes the combat tag on a blocked melee swing, which was the one thing that made #300 unsafe by itself, and the open-group question is answered in the #301 page (membership, not join policy). Still to clear before the set merges: #301 (two doc rows describe #303 and #304) and #304 (boot-time scan runs before mods patch).

Zaldaryon:
## Re-review requested

The root branch was rebased onto current `indev` after #301 and #304 merged. The README conflict resolution keeps both the role-prefix and friendly-fire command links.

Latest head: `3620e87`.

Verification: `make build` passed with 0 errors and 4 solution warnings. `bash scripts/smoke-test.sh` passed with 14 console commands and reached `WorldReady`. Please re-review the current head and the complete stack.

Pixnop (approved):
Re-checked 3620e87 against indev: the branch is exactly the reviewed stack, #300 plus the #301, #302 and #304 content I approved PR by PR (toggle, smoke probe, docs page, explosion filter, Harmony warning with both call sites, privacy fix), and the README resolution keeps both the role-prefix and the friendly-fire links. Nothing unreviewed appeared in the rebase. Approving the root on the new head; #303 is the last piece.

Pixnop:
Three small ones on the rebased root, none of them changes my approval, all of them fall out of the merge order. StratumHarmonyVisibility.cs:77-91 dropped ServerMain.CreateExplosion, which was right on #304's own base, but this head carries the #302 filter in CreateExplosion, so a mod that patches it and skips the original now defeats the blast rule with no warning naming it; worth putting the entry back. friendlyfire.md:10-12, 98 and 147 still say the Harmony warning lands with #304, and it is in this head. And once #303 goes in, the same lines describe the source seams as future work too.

Zaldaryon:
## Re-review requested

PR #303 was merged into this PR's base branch as `6766705`. The new #300 head is `6766705`, and its tree is identical to the reviewed #303 head `06defb5`; no source change was introduced beyond the approved stack.

The previous #300 approval references `3620e875`, so please re-review the current head before merge.

Zaldaryon:
## Fixes applied

- Added `ServerMain.CreateExplosion` to the friendly-fire Harmony watch list.
- Updated the friendly-fire documentation to describe the complete source-seam and explosion coverage.

## Verification

- `make build` passed with 0 errors.
- `bash scripts/smoke-test.sh` passed: 14 commands, `WorldReady`.

@Pixnop Please re-review the current head `f7ca5a4`.

Pixnop (approved):
Re-checked f7ca5a4. The #303 merge commit 6766705 has the same tree as the #303 head I approved (git diff is empty), and the follow-up commit only touches two files: ServerMain.CreateExplosion is back in the watch list at StratumHarmonyVisibility.cs:91, and the page now describes the seams, the explosion filter and the warning as present. Approving on this head.

One line to align before or after merge, your call: the message table at friendlyfire.md:55 now says "N blocked hit(s) since restart." while StratumFriendlyFireSystem.cs:186 still emits "N source hit(s) blocked since restart.", and the page presents that table as exact strings. The eight line offset in the ServerSystemEntitySimulation.cs.patch hunk headers is still there too, a regen through extract-patches.sh clears it.

Zaldaryon:
## Follow-up fixes

- Aligned the documented `/friendlyfire status` counter with the emitted `N source hit(s) blocked since restart.` message.
- Regenerated `ServerSystemEntitySimulation.cs.patch` so its hunk headers match the extracted source.

## Verification

- `make refresh` passed; all patches applied.
- `make build` passed with 0 errors in both regular and embedded builds.
- `bash scripts/smoke-test.sh` passed: 14 commands, `WorldReady`.

@Pixnop Please re-review the current head `2fafeb2`.

Pixnop (approved):
Re-checked 2fafeb2: two files, the message table at friendlyfire.md:55 now quotes the string the code emits, and the regenerated ServerSystemEntitySimulation.cs.patch carries the canonical headers (+1573, +1616, +1753, +1778) with the same applied content, only context attribution moved. Nothing else changed. Approving on this head; from my side the stack is ready to merge.

PR #301, opened 2026-09-07
Group friendly fire: docs and shared group check (1/4, part of #277)

Zaldaryon:
## Summary

First of four stacked follow-ups to #300. This PR documents the toggle and the group-membership rule, records the behavior delivered by #302, and fixes the map-pin privacy check for downgraded memberships.

Base branch is `feat/issue-277-group-friendly-fire` (#300). Merge #300 first, then this rebases onto `indev`.

### Change

- **`docs/commands/friendlyfire.md`** (new): documents the toggle, the on/off convention, the group rule, the exact status and toggle strings, the config keys, and the damage vectors that the current branch and follow-up PRs cover. The page identifies source-seam enforcement from #303 and Harmony warnings from #304 as stack follow-ups.
- **`README.md`**: links the new page next to the kits page.
- **`sources/VintagestoryLib/Vintagestory.Server/StratumPlayerPrivacy.cs`**: routes the map-pin group exception through the shared `StratumPlayerGroups.SharesGroup` instead of its own local copy. The local copy matched any shared `GroupUid`, including a membership downgraded to level `None`. That leaked exact coordinates under `AllowGroupMapVisibility` to a player who was no longer really in the group. The shared helper requires a live membership on both sides and reads the membership dictionaries directly instead of `IPlayer.Groups`.
- **`sources/VintagestoryLib/Vintagestory.Server/StratumFriendlyFireConfig.cs`**: documents the damage-over-time rule.
- **#302, merged into this branch**: filters known-igniter explosions so the igniter's group mates are not damaged. A bomb lit by an offline player may have no resolvable igniter.

### User-visible changes

- A player whose group membership is downgraded to `None` no longer receives exact map coordinates through the group visibility exception.
- When the explosion igniter is known, the friendly-fire toggle protects the igniter's group mates from that explosion.

### Gates

- `dotnet build VintageStory.slnx -c Release` then `... -p:EmbedPatchedFiles=true`: green, 4 pre-existing `NU1904` warnings, none on touched files.
- `bash scripts/smoke-test.sh`: `PASS: server reached RunGame, no fatal errors, 13 console command(s) verified.`
- Private Atlas suite (`research/atlas-tests/stratum-pr-validation/FriendlyFireScenarios.cs`): the `/friendlyfire` control-path scenarios pass. The scenarios that need a second connected player remain blocked by the 1.22.7 Atlas synthetic-join gap.
- `scripts/extract-patches.sh`: no patch or vanilla-source change. All edits are in `sources/` and docs.

### Limitations

- `StratumPlayerPrivacy` now excludes a `None`-level membership from the map-pin group exception. That is the intended visible behavior change.
- The source-seam guards and Harmony conflict warning land in #303 and #304.

## Type

- [ ] Bug fix
- [ ] Performance
- [ ] New feature
- [x] Refactor or cleanup
- [x] Docs or build

## Checklist

- [x] `scripts/extract-patches.sh` ran clean (no patch touched).
- [x] Every vanilla edit has a `// Stratum` marker (none in this PR).

## Related issues

Part of #277


Pixnop (commented):
The privacy cleanup is the good part here. Routing both call sites through StratumPlayerGroups.SharesGroup kills a real GroupUid-only match and drops two array allocations off a 100 ms per-pair tick, and the comment left at StratumPlayerPrivacy.cs:97 explains why the old copy was wrong instead of just deleting it. The page reads well.

Three things on the page. friendlyfire.md:79 says explosions are covered; StratumFriendlyFireConfig.cs:14, in the same commit, says they are not. #302 fixes the comment, nothing fixes the page if #302 slips. Lines 88-90 describe several enforcement points and the Harmony warning in the present tense; both land in #303 and #304. And the same sentence says the warning fires "while the toggle is on". The code in #304 warns when BlockGroupDamage is true, that is /friendlyfire off, which matches the convention your own page sets two paragraphs earlier. The #304 comment at StratumHarmonyVisibility.cs:74 has the same inversion.

While you are in the page: the "does not cover" list should name the vectors that really slip through, all of them block or AI sourced with no cause entity. Fire (BEBehaviorBurning, firepit, forge, coal pile), falling sand and gravel (EntityBlockFalling), creature and commanded-animal attacks (AiTaskMeleeAttack sets no CauseEntity), the bees a thrown beenade spawns, and the leave-group, hit, rejoin bypass that the deferred admin group commands were meant to close. None of that needs code in this batch, but an admin flipping the toggle should know.

Smaller: line 115 promises that /stratum access command friendlyfire shows the cooldown; FormatAccessDetails at CmdStratum.cs:473 prints Enabled and Privilege only (kits.md:247 has the same sentence, so it is copied, not new). And the consolidation stops halfway: SystemRemotePlayerTracking.cs.patch:121 still matches on raw GroupUid in the vanilla fallback, so with HideMapPins false a None-level member keeps a pin at unlimited range. Scope call?

Last one: commit 63db514 bundles the page, the README link and a user-visible privacy change under a docs label. One logical change per commit, and that fix deserves its own changelog line.

Pixnop (changes requested):
Switching to a change request for the page fixes listed above: the explosions and multi-point wording at lines 79 and 88-90, the on/off polarity of the Harmony warning, and the privacy fix pulled into its own commit with a changelog line. The rest are questions.

Zaldaryon:
## Re-review requested

Pixnop (commented):
Thanks, the page reads better. friendlyfire.md:97 now says the warning fires while friendly fire is off, which matches the convention, and #302 landing in the branch settles the explosions contradiction: StratumFriendlyFireConfig.cs:14 and the page agree. The no-cause list at 83-87 names the vectors that really slip through, and 122 no longer promises a cooldown that FormatAccessDetails never prints.

Two things left before I approve. The sync table at friendlyfire.md:133 now lists ServerSystemEntitySimulation.HandleEntityInteraction and EntityProjectileBase.CanDealDamage as blocking seams, and the warning clause at 96 still describes the Harmony warning as shipped. Grepping BlockGroupDamage on this head finds EntityPlayer.cs.patch:15 and ServerMain.cs.patch:1007 and nothing else, so those describe #303 and #304. A short "lands with #303 / #304" on each is enough, or merge the stack in order, but the page as it stands promises code this branch does not carry.

The other is the body. Since 1215c26 the branch also ships the explosion filter from #302, a real behaviour change to the toggle, while the description still says docs and cleanup only. The commit split I asked for is not worth it any more with #302 merged on top, so I withdraw that; a changelog line for the privacy fix and one for explosions in the body covers it.

Small one at 81: the uid survives a restart (BEBomb writes ignitedByPlayerUid), what fails is PlayerByUid for an igniter who is not online, restart or not. "Igniter offline" is shorter and right. SystemRemotePlayerTracking.cs.patch:121 stays open if you want it deferred, just say so.

Zaldaryon:
## Re-review requested

The documentation and PR description now match the current branch behavior. They identify the map-pin privacy change and #302 explosion filter already present here, and mark the source-seam and Harmony follow-ups as #303 and #304 work.

Latest commit: `5e6116f`.

Verification: `git diff --check` passed for the Markdown change. Please re-review the current head.

Pixnop (approved):
The page and the body now say what the branch carries. friendlyfire.md:10-12 and the sync table at 139 mark the source seams and the Harmony warning as #303 and #304 work, 87 says offline igniter instead of restart, and the body lists the map-pin privacy change and the #302 explosion filter under user-visible changes. Approving.

SystemRemotePlayerTracking.cs.patch:121 keeps the raw GroupUid match in the vanilla fallback; I take that as deferred, fine as a follow-up.

Pixnop:
Correction on one line I asked for. Your original wording at friendlyfire.md:87 was right and mine was wrong: on the server, PlayersByUid is only cleared by SaveGame on world load and DisconnectPlayer never prunes it, so PlayerByUid still resolves an igniter who logged off during the session. What actually loses the igniter is a fuse that outlives the process, so "lit before a restart" is the accurate phrase. Sorry for the detour. While you are there, 98-101 lost the subject for "bypasses this" when the #304 attribution went in front. Neither changes the approval.

PR #303, opened 2026-09-07
Group friendly fire: block at the source seams (3/4, part of #277)

Zaldaryon:
## Summary

Third of four stacked follow-ups to #300. Moves the group friendly fire block from the health check up to the two seams where an attack is actually decided, which fixes three of #300's stated limitations at once.

Base branch is `feat/issue-277-ff-explosions` (PR 2/4).

### Why

The #300 hook in `EntityPlayer.ShouldReceiveDamage` stops the damage but runs late. By the time it fires, a melee swing has already:

- played `sounds/player/slap` (`EntityAgent.OnInteract`),
- spent a point of weapon durability (`Collectible.OnAttackingWith` calls `DamageItem`),
- fired `OnPlayerInteractEntity`, which combat-tags both players (`StratumCombatLogSystem`).

A blocked shot similarly still paid the arrow its durability and impact sound in `ImpactOnEntity`.

### Change

- **`sources/VintagestoryApi/Server/StratumFriendlyFireHook.cs`**: adds `OnBlockedAttack`, an `Action<string,string>` the seams call so the notice logic lives in one place.
- **`sources/VintagestoryLib/Vintagestory.Server/StratumFriendlyFireGuard.cs`** (new): `BlocksMeleeAttack`, so the vanilla patch stays a one-line call.
- **`patches/VintagestoryLib/Vintagestory.Server/ServerSystemEntitySimulation.cs.patch`**: the guard call in `HandleEntityInteraction`, right after the vanilla PvP gate, only for attacks (`flag`), never right-click interactions like healing a group mate. Returning there is exactly the vanilla PvP-off path: no interact event, so no slap, no durability, no combat tag.
- **`patches/VSEssentials/Entities/EntityProjectileBase.cs.patch`**: the same rule inlined in `CanDealDamage` next to the existing `AllowPvP` check (VSEssentials cannot reference VintagestoryLib, so it uses the API-only types and fires the hook directly).
- **`sources/VintagestoryLib/Vintagestory.Server/StratumFriendlyFireSystem.cs`**: `OnBlockedAttack` points at a per-attacker throttled chat notice and a blocked-hit counter shown in `/friendlyfire status`.
- **`sources/VintagestoryLib/Vintagestory.Server/StratumFriendlyFireConfig.cs`**: `NotifyBlockedAttacker` (default true), `NotifyThrottleMs` (default 3000, clamped 500 to 60000), `BlockedMessage`. `StratumConfig` wires the `EnsureSane` call.
- **`docs/commands/friendlyfire.md`**: the new config keys and the status counter.

The `ShouldReceiveDamage` hook from #300 stays as the catch-all for mod-authored damage that calls `ReceiveDamage` normally, and as the test source for the explosion filter.

### No combat tag on a blocked hit

This inverts what #300's body asked for. Today a blocked melee swing wrongly tags both players (the projectile path already does not tag, gated on `didDamage`). A hit that deals no damage is not combat, and letting a group mate tag you by holding left-click is an exploit. The melee seam returning before `TriggerPlayerInteractEntity` fixes it for free, no change to `StratumCombatLogSystem`.

### Gates

- Two-pass build green, 4 pre-existing `NU1904`.
- `bash scripts/smoke-test.sh`: `PASS ... 13 console command(s) verified.`
- `scripts/extract-patches.sh`: `ServerSystemEntitySimulation.cs.patch` and `EntityProjectileBase.cs.patch` only. The SSE patch has one two-line brace reshuffle from the local decompiler in an unrelated adjacent hunk, otherwise the diff is the new melee hunk plus offset bumps. `ServerMain.cs.patch` from PR 2 is left untouched.
- Private Atlas suite: the control-path scenarios pass. The melee/projectile durability and no-tag scenarios need two connected players and are blocked by the same 1.22.7 synthetic-join gap as #274's suite.

### Limitations

- With friendly fire off, swinging at a group mate no longer plays the slap sound. Intended.
- A mod that Harmony-replaces `Entity.ReceiveDamage`, the interaction handler, or the projectile path and skips the original still bypasses this. PR 4/4 logs a warning naming such a mod. No server platform can hard-prevent it.
- Blocked hits stay out of the combat log entirely (deliberate); the `/friendlyfire status` counter is the visibility.

## Type

- [ ] Bug fix
- [ ] Performance
- [x] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

- [x] `scripts/extract-patches.sh` ran clean.

## Related issues

Part of #277


Pixnop (changes requested):
Moving the block up to the interaction handler is the right call. The return lands exactly where the vanilla PvP gate returns, so the slap sound, the durability tick and the combat tag go away for free and StratumCombatLogSystem needs no change. I rebuilt the patched file from both patch versions and the result is base plus the one guard hunk, so the brace reshuffle in the SSE patch is realignment only.

The blocker is threading. StratumFriendlyFireSystem.cs:24 says both fields are only touched from the main thread. The melee seam is. The projectile seam is not: EntityProjectileBase.cs:78 wires OnPhysicsTickCallback, BehaviorPassivePhysics.cs:389 invokes it from OnPhysicsTick, and PhysicsManager.DoWork runs OnPhysicsTick on the dedicated physics threads whenever tickables exceed Performance.Physics.ParallelThreshold. Our default is 32 (vanilla's is 800), so on any populated server the arrow path runs on a worker. From there HandleBlockedAttack does a plain Dictionary insert at line 73, a non-atomic long increment at line 60, and a SendMessage. Two arrows blocked in the same physics pass from two attackers can resize lastNotifyMs concurrently, and a torn Dictionary in .NET either throws or spins forever, on a physics thread. Rare, but the outcome is a physics stall, and the fix is small: ConcurrentDictionary plus Interlocked.Increment, or EnqueueMainThreadTask for the notice. StratumCombatLogSystem.OnProjectileHit has the same shape against activeTags, so that one is already exposed and could go in the same pass or a follow-up.

Smaller: lastNotifyMs is never pruned, not even on disconnect (StratumCombatLogSystem clears its own state there). friendlyfire.md:26 and the sync table at :129 still name ShouldReceiveDamage as the blocking point. And blockedHitsThisBoot only counts the two new seams, so /friendlyfire status stays at zero after a bomb blast that #302 filtered, while the doc row presents it as the count of blocked hits.

One question: string.Format on cfg.BlockedMessage at line 80 throws on a stray brace, now from inside a physics tick, and EnsureSane only rejects blank. Same exposure as TagMessage, so this is about the house convention rather than this PR.

Zaldaryon:
## Re-review requested

Pixnop (approved):
Thanks, the threading fix reads clean. ConcurrentDictionary at StratumFriendlyFireSystem.cs:27, Interlocked.Increment at 62 and the EnqueueMainThreadTask hop at 77 put the player lookup and SendMessage back on the server thread, TryClaimNotification holds up when two workers claim the same uid, and the FormatException catch at 122 answers the format question. Approving.

Two follow-ups I would rather see in an issue than block on. CanDealDamage still calls StratumPlayerGroups.SharesGroup on the physics worker, and that enumerates ServerPlayerData.PlayerGroupMemberships (StratumPlayerGroups.cs:43) while /group mutates it on the main thread; same hazard, narrower window, and the vanilla PvP check on the same line already reads player data from that thread, so this is a pre-existing class rather than something this PR adds. StratumCombatLogSystem.OnProjectileHit has the same shape against activeTags. Both fit one issue.

Smaller: friendlyfire.md:26 and the intro at 5-9 still name ShouldReceiveDamage as the blocking point, only the sync table moved. lastNotifyMs still has no disconnect hook, the TryRemove at 112 fires only when a claim outlives the attacker. The catch at 122 logs nothing, so a mistyped BlockedMessage is invisible for good; one warning on first failure would do. And "source hit" at 185 is a term the page never defines.

Zaldaryon:
## Re-review requested

The branch was rebased onto the corrected #301 documentation and now targets the #300 stack root. The source changes are unchanged from the previously approved implementation.

Latest head: `44b7683`.

Verification: `make build` passed with 0 errors and 4 solution warnings. `bash scripts/smoke-test.sh` passed with 13 console commands and reached `WorldReady`. Please review the current head and base.

Zaldaryon:
## Re-review requested

The branch was rebased onto the current #300 root at `3620e87`. The conflict resolution preserves the root reach guards and places the friendly-fire melee guard before the interaction event. The nested patch format was refreshed and verified from a clean bootstrap.

Latest head: `06defb5`.

Verification: `make refresh` applied every patch, `make build` passed with 0 errors and 4 solution warnings, and `bash scripts/smoke-test.sh` passed with 14 console commands and reached `WorldReady`. Please review the current head and base.

Pixnop (approved):
Re-checked 06defb5 against the new root: the guard, hook, projectile hunk and system are byte for byte what I approved, plus the WarnFriendlyFireConflicts line inherited from #304. The melee seam landed in the right place after the rebase, behind StratumEntityReachGuard and the dimension check from indev and right after the vanilla PvP-off return, so the reach guards run first and a blocked group hit still skips the interact event. Approving on the new head.

Pixnop:
One thing from the rebase, not blocking. The inner hunk of ServerSystemEntitySimulation.cs.patch grew by the eight guard lines but the four hunk headers after it kept their old new-side starts, so git apply lands hunks 9 to 12 at an eight line offset; a regen through extract-patches.sh gives +1616, +1753, +1778 and +1803 and makes the file canonical again. Same doc note as on #300: friendlyfire.md:10-12 describes these seams as future work on the branch that adds them.

PR #308, opened 2026-09-09
Repository hygiene: SDK pin, editorconfig, owners, license detection, docs index

Pixnop:
## Summary

Repository hygiene, eight small commits, no code change. Each one is independent and can be dropped on its own.

- **global.json and .editorconfig.** The SDK is pinned to the 10.0 line with `rollForward: latestFeature`, the same major the release workflow installs, so a contributor on a different SDK gets told at restore time instead of at review time. The editorconfig writes down what CONTRIBUTING already asks for: tabs in Stratum-owned C# and in the decompiled VintagestoryLib tree, four spaces in the Anego forks, two spaces in project and YAML files. `*.patch` is excluded from whitespace trimming so an editor cannot corrupt a unified diff.
- **CODEOWNERS by login.** The file pointed at `@StratumServer/core` and `@StratumServer/docs`, and the org has no teams, so it matched nobody and GitHub's validator reported four unknown owners. It now lists the admins, plus Zaldaryon on docs. It is a suggestion list, nothing requires a code owner review; add yourself where you want the pings.
- **SECURITY.md link.** Private reports went to `github.com/trevorftp/Stratum/security/advisories/new`, a personal fork. Now the org repository.
- **NOTICE.md in releases.** `scripts/pack-release.ps1` stages it next to LICENSE and README.md, so the README's link resolves inside every archive.
- **LICENSE detection.** GitHub reports the license as "Other" because the scope paragraph sits inside the MIT text. The paragraph moves verbatim to `NOTICE.md`, the README license section points at it, and LICENSE is the plain MIT text again. Terms unchanged. This is the one commit I would understand someone wanting to keep out; say so and I drop it.
- **Dependabot groups.** One weekly pull request for all action bumps instead of one per action. NuGet stays out on purpose: the server references the package versions the game ships, and a bump there breaks the overlay against vanilla.
- **docs/README.md.** An index of the four pages by audience, linked from the README.
- **FUNDING.yml** trimmed to the platform in use.
- **CODE_OF_CONDUCT.md.** Contributor Covenant 2.1 by reference, with the reporting path SECURITY.md already uses.

## Type

- [ ] Bug fix
- [ ] Performance
- [ ] New feature
- [ ] Refactor or cleanup
- [x] Docs or build

## Checklist

- n/a `scripts/extract-patches.sh`: nothing under `patches/` changes.
- n/a Build: no source changes. `dotnet --version` resolves 10.0.111 through the new global.json.
- n/a `// Stratum` markers: no vanilla file changes.
- n/a Server start: nothing in this PR runs.

## Branch protection follow-up

Today `indev` has no ruleset while it is the target of every pull request, and `main` is protected only by the "Restrict Main" ruleset. Once this PR is merged I will add a ruleset on `indev` through the repository settings, no file involved:

- Name `Protect indev`, target `refs/heads/indev`, enforcement active.
- Block branch deletion and force pushes.
- Require a pull request with one approving review, stale reviews dismissed on push, no code owner review required, merge, squash and rebase all allowed.
- No bypass actor. Admins included, so the rule applies to me too.
- No required status check yet. The CI workflow that builds on pull requests is coming in a separate PR, and the check gets added to the ruleset once it has run green on a few PRs.

Repository security settings (secret scanning with push protection, Dependabot alerts and security updates) are off today; I will turn them on at the same time, they change nothing in the tree.

## Related issues

Part of the housekeeping around #147.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,47 @@
+# Editor defaults for the tracked tree. Mirrors CONTRIBUTING.md: tabs in Stratum
+# C#, match the surrounding style when editing vanilla, LF everywhere except
+# PowerShell (see .gitattributes).
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+# Stratum-owned C#: tabs, file-scoped namespaces (enforced by review, not here).
+[{sources,StratumServer,tests}/**.cs]
+indent_style = tab
+
+# Decompiled VintagestoryLib is emitted with tabs by ilspycmd.
+[VintagestoryLib/**.cs]
+indent_style = tab
+
+# The Anego forks (API, Essentials, Survival) use four spaces. Keep them as they are.
+[{VintagestoryApi,VSEssentials,VSSurvivalMod}/**.cs]
+indent_style = space
+indent_size = 4
+
+# Unified diffs must keep their trailing whitespace and context lines byte for byte.
+[*.patch]
+trim_trailing_whitespace = false
+insert_final_newline = false
+
+[*.{csproj,props,targets,slnx,json,yml,yaml}]
+indent_style = space
+indent_size = 2
+
+[*.md]
+trim_trailing_whitespace = false
+
+[*.sh]
+indent_style = space
+indent_size = 2
+
+[*.ps1]
+end_of_line = crlf
+indent_style = space
+indent_size = 4
+
+[Makefile]
+indent_style = tab
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,24 +10,34 @@ insert_final_newline = true
 trim_trailing_whitespace = true
 
 # Stratum-owned C#: tabs, file-scoped namespaces (enforced by review, not here).
+# A few older files under sources/ are still space-indented; they are the ones
+# out of line, and reformatting them is a separate cleanup.
 [{sources,StratumServer,tests}/**.cs]
 indent_style = tab
 
-# Decompiled VintagestoryLib is emitted with tabs by ilspycmd.
-[VintagestoryLib/**.cs]
+# Generated working trees. extract-patches diffs them against the pristine
+# baseline, so an editor that trims or reformats one of these files on save
+# turns into real hunks in patches/. Leave their whitespace alone.
+#
+# Decompiled by bootstrap into baseline/, tab-indented by ilspycmd.
+[baseline/{VintagestoryLib,VintagestoryServer}/**.cs]
 indent_style = tab
+trim_trailing_whitespace = false
+insert_final_newline = false
 
-# The Anego forks (API, Essentials, Survival) use four spaces. Keep them as they are.
-[{VintagestoryApi,VSEssentials,VSSurvivalMod}/**.cs]
+# Anego forks cloned by bootstrap (forks.json), four spaces.
+[{VintagestoryApi,Cairo,VSCreativeMod,VSEssentials,VSSurvivalMod}/**.cs]
 indent_style = space
 indent_size = 4
+trim_trailing_whitespace = false
+insert_final_newline = false
 
 # Unified diffs must keep their trailing whitespace and context lines byte for byte.
 [*.patch]
 trim_trailing_whitespace = false
 insert_final_newline = false
 
-[*.{csproj,props,targets,slnx,json,yml,yaml}]
+[*.{csproj,props,slnx,json,yml,yaml}]
 indent_style = space
 indent_size = 2
 
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,4 +1,7 @@
-* @StratumServer/core
-/docs/ @StratumServer/docs
-.github/ @StratumServer/core
-/scripts/ @StratumServer/core
+# Review suggestions, not a gate: the indev ruleset does not require a code
+# owner review. The org has no GitHub teams yet, so owners are listed by login.
+# Add yourself when you take an area.
+*           @trevorftp @pizza2004 @Pixnop
+/docs/      @trevorftp @pizza2004 @Pixnop @Zaldaryon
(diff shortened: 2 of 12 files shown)


Zaldaryon (changes requested):
The CODEOWNERS replacement is the reason to take this PR. `@StratumServer/core` and `@StratumServer/docs` do not exist, the org has no teams at all, and GitHub's validator reports four unknown-owner errors against `indev` right now, so code ownership is silently off for every path. Every login you list has push access, so the new file validates clean. The LICENSE change is fine too: the seven removed lines are a project scope note, not part of the MIT text, and `NOTICE.md` carries them forward word for word. What ships in `LICENSE` now is the unmodified MIT body, which is also what makes GitHub's license detection work.

Two things need a fix before I approve.

**Release archives get a dead link.** `scripts/pack-release.ps1` stages exactly `LICENSE` and `README.md` into each zip. The README you ship now ends its license section with a link to `NOTICE.md`, and that file is not in the archive, so every release gets a broken link and the Anego scope note stops traveling with the artifacts. Add `NOTICE.md` to the staging copy after line 98 and to the file list in the header comment at line 8.

**The editorconfig rule for the decompiled tree matches nothing.** `bootstrap.sh` decompiles `VintagestoryLib` and `VintagestoryServer` into `baseline/`, and that is the path the solution and the patch step use. Your `[VintagestoryLib/**.cs]` section points at a directory that never exists, and `VintagestoryServer` gets no rule at all. Use `[baseline/{VintagestoryLib,VintagestoryServer}/**.cs]`. While you are there, take that tree and the fork working directories out of `trim_trailing_whitespace`: `extract-patches` diffs the working tree against the baseline, so an editor that trims a decompiled file on save turns into real hunks in `patches/`, which the "What not to patch" section tells contributors to drop.

Smaller items for the same pass:

- `Directory.Build.targets` is tab-indented top to bottom, so listing `targets` in the two-space group fights the only file it matches. Drop `targets` from that group.
- The fork group covers three of the five entries in `forks.json`. `Cairo` and `VSCreativeMod` are missing. Nothing patches either today so it is harmless for now, but add them once you have checked their indent.
- The `sources/` rule itself is right, tabs matching CONTRIBUTING. It will flag a handful of space-indented files already under `sources/`. Those files are the ones out of line, and reformatting them is a separate cleanup, so just expect the noise.

Nits. The CODEOWNERS comment says the "indev ruleset" does not require a code-owner review, but `indev` has no ruleset and no protection at all, so reword it to "no branch rule requires it". `CODE_OF_CONDUCT.md` uses "behaviour" and "judgement" where the repo is mostly American, and nothing links to it, not even the new `docs/README.md`.

One repo-side item, not yours: private vulnerability reporting is turned off, so the advisory link in `SECURITY.md` returns 404 for anyone without write access. It needs turning on in repo settings before this merges.

I left `global.json` and `dependabot.yml` alone. `10.0.100` with `latestFeature` resolves against the `10.0.x` that `release.yml` installs, and the grouped-actions block is valid v2. Skipping NuGet is the right call given the versions the server pins.

The checklist boxes about `extract-patches`, the build and the smoke test do not apply here. Nothing in this PR touches `patches/` or compiles, so treat them as not applicable rather than unchecked.

Pixnop:
Thanks, every point checked out, and the three new commits cover them.

- `NOTICE.md` now ships in every release archive: `scripts/pack-release.ps1` lists it in the header and stages it after `README.md`, so the README link resolves inside the zip.
- The editorconfig section now targets `baseline/{VintagestoryLib,VintagestoryServer}`, and that tree plus the five fork directories from `forks.json` keep their whitespace (no trim, no forced final newline), so a save in an editor cannot turn into a hunk in `patches/`. Cairo and VSCreativeMod joined the four-space group after checking their indent, and `targets` left the two-space group.
- The CODEOWNERS note no longer names a ruleset, the code of conduct uses the repo's American spelling, and the docs index links to it.
- The checklist items that do not apply are marked n/a in the body.

Private vulnerability reporting is on in the repository settings now, so the advisory link in `SECURITY.md` resolves for anyone.

Zaldaryon (approved):
Both blockers are fixed, and the rest of the list closed out with them.

`NOTICE.md` now ships in release archives. `pack-release.ps1` stages it next to `LICENSE` and `README.md` (the copy sits right after the other two, and the header comment lists it), so the README's link resolves inside the zip instead of pointing at a file that isn't there.

The `.editorconfig` paths match real trees now. `[baseline/{VintagestoryLib,VintagestoryServer}/**.cs]` lines up with what `bootstrap.ps1` actually decompiles into, and the fork group picks up `Cairo` and `VSCreativeMod` alongside the other three `forks.json` entries. `targets` came out of the two-space group, so it stops fighting the tab-indented `Directory.Build.targets`, and the sixteen files still left in that group are genuinely two-space, so narrowing it didn't break anything else. The `sources/` comment about the seven space-indented holdouts reads accurately.

CODEOWNERS no longer names a ruleset that doesn't exist. Today `indev` has no branch rule at all, and the only active ruleset in the repo is `Restrict Main`, scoped to the default branch. `CODE_OF_CONDUCT.md` is linked from `docs/README.md` and reads American spelling throughout, though `docs/role-prefixes.md` and `docs/commands/friendlyfire.md` still use British spelling in a few places. Pre-existing, outside this PR, just noting it for whoever picks up docs consistency next.

Private vulnerability reporting is on, so the `SECURITY.md` advisory link resolves for non-write users now. That was the one item on my side, and it's done.

Two small things worth a look sometime, neither blocking: `.targets` has no explicit `indent_style` anymore and falls through to the `[*]` default, so an explicit tab setting would be more precise than implicit. And `.baseline`, the pristine snapshot side `extract-patches` diffs against, isn't exempted from whitespace trimming the way the two working-tree sides are. Low risk since nobody edits it by hand.

Approving.

Pixnop:
The `Protect indev` ruleset is now active, as described above: no force pushes or deletion on `indev`, and changes land through a pull request with one approving review. No bypass, admins included, and no required status check until the CI workflow has a few green runs behind it.

PR #309, opened 2026-09-09
Add the Atlas scenario suite (tests/StratumScenarios)

Pixnop:
## Summary

Adds `tests/StratumScenarios`, seventeen end-to-end scenarios that boot a real server in-process through the [Atlas](https://github.com/Pixnop/Atlas) harness and assert behaviour this fork documents: chunk persistence across save, unload and reload; the simulation distance throttles for entities, random ticks and block tick listeners; the fact that each `stratum-performance.json` toggle restores vanilla behaviour; and that `/stratum` and the homes command group are registered at all. `BootScenarios` also guards the worst false green this repo can produce: a build made without `-p:EmbedPatchedFiles=true` boots the downloaded vanilla lib, and every other scenario would pass while testing vanilla.

This is the first half of #147. The scenarios are transposed from StratumParity, where they have run against every stable release since July, with the vanilla branches removed: each probe became an absolute assertion on Stratum behaviour, and the margins, anchoring and timeouts did not change. The vanilla comparison, the dashboard and the tick cost probes stay in StratumParity. The workflow that runs this on pull requests comes in a second PR, so that it lands unrequired and only becomes a required check once it has been green a few times.

### What runs

`make scenarios` (or `bash scripts/scenarios.sh`, or `.\scripts\scenarios.ps1` on Windows) builds, materialises the install with one `--stratum-prepare-only` launch (the launcher writes vanilla plus the patched overlay into its own output directory, which is not configurable, so that is what `VINTAGE_STORY` points at), and runs `dotnet test`. The project sits outside `VintageStory.slnx` on purpose: it needs a prepared install, and a normal build must not depend on one. Nothing runs it automatically. It pulls one NuGet package, `Pixnop.Atlas.XUnit` 0.13.1, which is the floor: 0.13.0 opened the synthetic handshake with the identification packet and the first-packet gate from c7c83f6 drops that, which is also why StratumParity has been red on indev since August 26.

### The one change outside `tests/`

`StratumServer/Program.cs`: `--stratum-prepare-only` now exits 1 when the build carries no embedded patched files, and refuses `--stratum-skip-bootstrap` with exit 2 since that pair prepares nothing. It already printed the warning and returned 0, and `Apply` writes nothing in that case, so on a repeat run the install keeps the patched lib an earlier build put there and looks prepared. With the non-zero exit, `make scenarios` stops before `dotnet test` instead of measuring yesterday's lib. A normal launch keeps warning and continuing.

### Numbers

- 17 scenarios in 8 classes, one server boot per class.
- 1 min 28 s to 1 min 33 s reported by `dotnet test` on the prepared indev install, zero build warnings; about 2 min 30 s on a slower machine.
- Diff: 24 files, about 950 lines, of which the six fixture files and the probe mod are copies.

## Type

- [ ] Bug fix
- [ ] Performance
- [x] New feature
- [ ] Refactor or cleanup
- [x] Docs or build

## Checklist

- [x] `scripts/extract-patches.sh`: no patch belongs to this PR. Running it reflows `ServerMain.cs.patch`, the hand-spliced form that came with #302, into canonical hunks with the same applied content; I left that file alone so the diff stays about the suite.
- [x] `dotnet build VintageStory.slnx -c Release -p:EmbedPatchedFiles=true` is green, 0 warnings.
- [x] Every vanilla edit has a `// Stratum` marker (none in this PR; `Program.cs` is Stratum code).
- [x] Tested on a real server start: `bash scripts/smoke-test.sh` passed, 14 console commands verified, WorldReady; `make scenarios` 17/17.

## Related issues

First half of #147.


Zaldaryon (changes requested):
This is real work and it passes on a clean bootstrap: I built the branch from scratch and ran `make scenarios`, 17 passed and 0 failed, test phase about two and a half minutes. The scenarios assert the right things. The entity probes check exact tick counts against the engine's own simulation counter rather than a loose band, the far random-tick platform asserts exactly zero conversions with a written reason why zero is correct, and the persistence scenarios round-trip chunk moddata next to the blocks so a reload cannot pass on a still-generated column. `Server_Should_RunPatchedLib_When_Built` is the one I care about most, because a suite that quietly tested vanilla would be worse than none. Style is clean: tabs, file-scoped namespaces, no stray annotations, zero warnings from the test project.

Two changes before I approve.

**`make scenarios` has no Windows path, and item 3 of the checklist now makes it required.** Any change to chunk IO, entity ticking, or the simulation distance limits has to run it, and the parenthetical says to use Git Bash or WSL because there is no `.ps1`. Every other script in `scripts/` ships as a `.sh` and `.ps1` pair, and item 2 right above yours names both. A Windows contributor with only PowerShell cannot run a step the checklist calls required. This is the gap you raised on #306, so it should hold here too. Add `scripts/scenarios.sh` and `scripts/scenarios.ps1`, have the make target call the shell script the way `smoke` calls `smoke-test.sh`, and name both in CONTRIBUTING.

**The disabled-scenario fixtures rest on a claim `indev` already contradicts.** `EntityTickingDisabledScenarios.cs:14`, `BlockTickListenerDisabledScenarios.cs:10` and `RandomTickDisabledScenarios.cs:11` all say `LoadOrCreateConfig` reads `stratum-performance.json` only when `stratum.json` exists. `StratumRuntime.cs` loads the sidecar under a bare `if (performanceConfigExisted)` now, with a `// Stratum:` comment recording that exact fix. So the three `stratum.json` fixtures are not needed, and pinning them to `"ConfigVersion": 1` has a cost: `StratumConfigMigration.CurrentVersion` is 3, so every fixture-seeded boot runs the version 1 to 3 migration. My run log shows it: `[Stratum] config ready: upgraded config from version 1 to 3`. Three of the eight boots exercise a migrating config instead of a current one, and a later migration that touches `Performance` will change what these fixtures mean with nobody editing them. Set the `stratum.json` fixtures to `"ConfigVersion": 3` and fix the three comments.

One I would like fixed but will not block on. `ChunkPersistenceScenarios.cs:110` says the write pattern stays in the anchor's chunk column, and `:119` does not guarantee it: an 8 by 16 footprint anchored at `Spawn + (200,1,0)` stays in one column only while `Spawn.X % 32` and `Spawn.Z % 32` are small. It holds on the default 1024000 map, which is why this passes today, but on a world with different spawn alignment part of the pattern lands in the neighbouring column, which `SaveUnloadReload` never unloads, so those assertions read out of memory and prove nothing. `PlacePlatforms` already derives a safe corner from `(chunkX * 32) + 8`. Same trick here.

Smaller, all optional:

- The new `if (prepareOnly)` guard in `Program.cs:64` is inside the `if (!skipBootstrap)` block, so `--stratum-prepare-only --stratum-skip-bootstrap` still prints "prepare complete" and exits 0 with nothing embedded. Nothing passes both flags today, so it is latent, but the flag table in `README.md:97` should mention the new nonzero exit.
- `SERVER_DIR` in the Makefile hardcodes `net10.0` where `FrameworkVersion` is already in `Directory.Build.props`.
- The README says the run takes about a minute and a half. I saw two and a half for the test phase alone, before the two builds and the prepare launch.
- The two command-registration scenarios assert `NotEqual("nosuchcommand")`, which passes on any other error. `ChatCommands.Get("sethome") != null` asserts the actual contract.
- `TickCounterBehavior` can be `internal`. Only the test classes need to be public.
- The test project turns on `Nullable` and `ImplicitUsings` where the rest of the repo does not. Defensible for a standalone project, but one line in the README would keep the next reader from filing it as a style break.

On the parts I read rather than tested: `RegisterGameTickListener`, `SidedProperties.Behaviors.Add` and `SetModdata` plus `MarkModified` are all called from the scenario body, and all mutate collections the tick loop walks. If Atlas marshals scenario bodies onto the server thread, say so in the README. If it does not, the CI crash you describe in `AssemblyInfo.cs` may be that rather than the probe mod's asset build.

The new dev dependency on `Pixnop.Atlas.XUnit` 0.13.1 is a maintainer call, not a code issue, but worth making on purpose now that `make scenarios` is a documented gate: it is a 0.x package on a personal account, and the repo pins no NuGet source and keeps no lock file.

Pixnop:
Thanks for building it from scratch, that is the run that matters. Twelve commits on top, one per point.

- **Windows path.** `scripts/scenarios.sh` and `scripts/scenarios.ps1` do the same steps in the same order: a two-pass build when the launcher is missing, one `--stratum-prepare-only` that stops the run on a non-zero exit, `VINTAGE_STORY` set for `dotnet test`, extra arguments passed through, exit code propagated. `make scenarios` calls the shell script the way `smoke` does and forwards `CONFIGURATION`, and CONTRIBUTING names both scripts. I could not run the `.ps1` here, there is no PowerShell on this machine; it mirrors `smoke-test.ps1` and checks `$LASTEXITCODE` after every native call, so a Windows run would be welcome.
- **Fixtures.** You are right that indev reads the sidecar on its own now, and the three comments say so. I kept the `stratum.json` seeds, at `"ConfigVersion": 3`, because removing them does not avoid a migration: with only `stratum-performance.json` present, `StratumConfigMigration.GetLoadedVersion` returns 2 (line 38), so every seeded boot would log "upgraded config from version 2 to 3" instead. With the seed at the current version the three Disabled classes log "config ready: loaded config" and no boot in the run logs an upgrade. The comments now say the seed has to track `CurrentVersion`.
- **ChunkPersistence** writes from `(chunkX * 32) + 8, anchor.Y, (chunkZ * 32) + 8`, the same trick as `PlacePlatforms`, so the whole pattern stays in the column `SaveUnloadReload` unloads whatever the spawn alignment.
- **Registration scenarios** assert `ChatCommands.Get("stratum")` and `Get("sethome")` directly and no longer run the commands.
- **Launcher.** `--stratum-prepare-only` with `--stratum-skip-bootstrap` is refused with exit 2 before any side effect, the usage text and the flag table say what prepare-only returns, and the skip-bootstrap hint no longer shows in that mode.
- `TickCounterBehavior` is internal, and the suite README gives one and a half to two and a half minutes of test time plus the builds, notes the nullable context, and says when the scripts rebuild.
- **Threading.** Atlas runs every scenario body on the server's game thread: its `GameThreadScheduler` is the synchronization context the server pump drains between ticks, so `RegisterGameTickListener`, `Behaviors.Add` and `SetModdata` never race the tick loop. What can still overlap is work the engine starts off thread by itself, the server assets packet build at boot being the one that bit on CI, and that is what the `AssemblyInfo.cs` comment now says.

On the dependency, I am making that call on purpose. I maintain Atlas, it is MIT, the csproj pins 0.13.1, StratumParity has run on it since July, and it only ever touches this test project, which sits outside the solution. Nothing that ships in a release depends on it.

On the new head: 17/17 with 1 min 28 s of test time, and the smoke test still passes with 14 commands.

Zaldaryon (approved):
Both blockers are fixed, and I confirmed the second one empirically, not just by reading the diff.

`scripts/scenarios.sh` and `scripts/scenarios.ps1` now exist as a matching pair, same file modes as the rest of `scripts/`. Both read `CONFIGURATION` from the environment with a Release default, pull `FrameworkVersion` from `Directory.Build.props` instead of hardcoding `net10.0`, build only when the launcher is missing, run one `--stratum-prepare-only --stratum-no-banner` launch, then `dotnet test` with `VINTAGE_STORY` pointed at the prepared install and extra arguments forwarded. I forwarded an argument for real to check that last part: `bash scripts/scenarios.sh --filter "FullyQualifiedName~DisabledScenarios"` ran 3 of the 17 and passed. The Makefile target now calls the shell script instead of inlining its own copy, and the hardcoded `SERVER_DIR` variable is gone with it. CONTRIBUTING names both scripts and no longer says there's no Windows path.

The fixture `ConfigVersion` is 3, matching `StratumConfigMigration.CurrentVersion`, which is still 3. I ran the full suite against the new fixtures and grepped every server log the run produced: zero occurrences of "upgraded config from version" anywhere, where yesterday's run logged the 1-to-3 migration on every seeded boot. The three disabled-toggle classes each log "loaded config" now, not "upgraded config" or "created default config". The commit's secondary claim, that a data directory holding only `stratum-performance.json` reports as version 2 and still migrates, checks out against `StratumConfigMigration.cs:38` too.

Full suite: 17 passed, 0 failed, 0 skipped, bootstrap plus scenarios about six and a half minutes cold (one vanilla re-download in there) and roughly three minutes warm. No scenario got dropped in the cleanup.

The rest of the list is addressed the same way. The chunk persistence pattern's corner math keeps the write footprint inside the same 32-wide column `LoadColumn` and `SaveUnloadReload` use, regardless of where spawn lands in its chunk. The two command scenarios now assert through `ChatCommands.Get` instead of the weaker `NotEqual` check. The prepare-only-plus-skip-bootstrap refusal sits before the bootstrap block and exits 2, confirmed by running it. `TickCounterBehavior` is internal. The README documents both nonzero exits, the nullable and implicit-usings choice, and the thread-affinity guarantee for the four APIs that needed it, and I checked that guarantee against what Atlas 0.13.1 actually does (`RunOnGameThreadAsync`, `GameThreadScheduler.DrainPending`) rather than taking the prose on faith. `-LiteralPath` is in all three PowerShell call sites that needed it, and it held up against a scratch path with brackets in it.

Three small things, none blocking. The AssemblyInfo comment about game-thread execution reads ambiguously on a first pass, a clause or comma would fix it. The README's stated suite duration, a minute and a half to two and a half minutes, sits about twenty seconds under what I actually measured. And nothing in the repo ties the fixtures' `ConfigVersion` to `StratumConfigMigration.CurrentVersion` beyond a comment, so a future version bump would silently start migrating them again with the suite still green. Worth a one-line note somewhere, not worth another round now.

The `Pixnop.Atlas.XUnit` dependency question from the last round stands as is, nothing here changed it: still 0.13.1 off a personal account, still no source pin or lock file, still behind the documented gate in CONTRIBUTING. That's a call for whoever owns dependency policy, not something I'm blocking on.

Approving.

PR #313, opened 2026-09-11
Fix vanished player privacy leaks

### LINKED ISSUE #312: Vanished players still leak their position to other clients
Clients are vanilla, but a modified client can read everything the server sends it. `/vanish` (#213) hides the player at the tracking layer, so entity ESP does not see a vanished player. A client that reads packets still can locate them, through the channels below. This came up on the official Discord.

### Already covered

`/vanish` works at the tracking layer, not just in rendering. `ShouldHideEntityFromClient` gates `UpdateTrackedEntityLists` and `PrepareEntitySpawns` (`patches/.../PhysicsManager.cs.patch:432, 448, 1151`), and toggling it on sends an explicit despawn (`StratumStaffCommandState.cs:145-169`). Spawns, PlayerData, bulk positions, the UDP relay of player positions (`.baseline/.../ServerUdpNetwork.cs:204`, gated on `TrackedEntities`), the map pin (`StratumPlayerPrivacy.cs:87-95`) and `/near` all respect it.

### Spectator mode

Vanilla spectator mode is still a tracked player, so any entity ESP sees it. That is vanilla behavior and out of scope here: on a Stratum server, staff who need to stay unseen should use `/vanish`. Two spectator leaks are Stratum's own and cheap to close. `/near` lists spectators because it only skips vanished players (`CmdStratumStaffCommands.cs:444-448`), and `stratum.near` is a default player privilege. With `PlayerPrivacy.Enabled`, the spectator branch of `SystemRemotePlayerTracking.cs.patch:64-86` sends exact map pins that vanilla never sends.

### What gets around the vanish filter

- **Mounts.** The boat or elk is not an `EntityPlayer`, so it stays tracked. Its positions are relayed (`.baseline/.../ServerUdpNetwork.cs:286-291`) and `seatdata` carries the rider's EntityId (`EntityBehaviorSeatable.cs:298`): continuous exact position.
- **Teleport packet.** `EntityPlayer.cs:1534` calls `BroadcastEntityPacket` (`.baseline/.../ServerMain.cs:3976`), which reaches every connected client at any distance with the destination XYZ. That covers `/tp`, `/back`, `/home`, `/spawn`, jail and respawn.
- **Thread 0 bucket.** The filter `continue`s but leaves the entity in `threadedTrackedEntities[0]` (`PhysicsManager.cs.patch:432`), and `SendAttributesViaTCP` walks that list (`.baseline/.../PhysicsManager.cs:685`). It is the only list in serial physics, which covers any server under the 32-tickable parallel threshold, and on larger servers it covers whoever lands in the first thread's slice. Clients within tracking range keep receiving watched attribute updates keyed by the vanished player's EntityId, which works as a proximity oracle, and a full sync sends the whole entity including its position.
- **Reconnect.** Vanish state is cleared on disconnect (`CmdStratumStaffCommands.cs:1921`, `StratumStaffCommandState.cs:334`), so the player rejoins visible, with a join message, until they type `/vanish` again.
- **Side effects.** `SendPrioritySpawn` is unfiltered (`.baseline/.../PhysicsManager.cs:993`), so an arrow or a thrown spear spawns next to the shooter and carries `firedBy`. Breaking or placing a block broadcasts the block update to every client and the particles to clients that have the chunk, which puts the player within reach of that block, and outside creative the break sound goes to everyone within 32 blocks. Spectators cannot break or place blocks, so this one is vanish only.
- **Hysteresis.** A viewer who loses the right to see vanished players mid-session, outside `/vanish hideothers`, keeps them tracked within 1.1 times the range (`PhysicsManager.cs.patch:466-480`).

### Proposal

Keep one "hidden from this client" predicate and apply it wherever an entity reaches a client:

1. Filter `threadedTrackedEntities` as well as `TrackedEntities`.
2. Treat a mount with a hidden passenger as hidden, and reveal it on dismount.
3. Send the Teleport packet only to clients tracking the entity. Others have nothing to apply it to, but the client handler needs checking first.
4. Persist vanish across reconnects and apply it before the first spawn and join message.
5. Exclude spectators from `/near` and from the privacy hook.
6. Filter `SendPrioritySpawn` by the shooter. Sounds, particles and block updates are harder to hide without breaking gameplay, so I'd document them as a known limit.

Role and config changes should trigger the same refresh as `hideothers`, which closes the hysteresis case.

Zaldaryon:
## Summary

Apply the Stratum visibility rule to every tracked-entity and player-data boundary that can disclose a vanished player's position.

## Issue

Fixes #312.

## Root cause

The existing `/vanish` filter covered the main tracking path but left several disclosure paths outside the same predicate. Thread 0, hysteresis, position and animation updates, priority spawns, entity broadcasts, player-data broadcasts, mounts, reconnect state, `/near`, and spectator map pins could bypass the intended visibility decision.

## Changes

- Extend the visibility predicate to mounts carrying a vanished player while keeping mounts with ordinary passengers visible.
- Document shared mounts as a protocol limitation because visible seat data can still disclose the vanished passenger.
- Filter thread 0, hysteresis, state and non-state position updates, animations, tags, priority spawns, entity broadcasts, death packets, and death announcements.
- Filter player data during join and broadcast paths.
- Persist vanish state in server player data and restore it before the first join broadcasts.
- Exclude spectators from `/near` results and third-party map-pin disclosure.
- Keep the existing staff visibility and `hideothers` behavior in the shared predicate.
- Add a packet-level reconnect assertion and update the vanish documentation.

## Type

- [x] Bug fix
- [ ] Performance
- [ ] New feature
- [ ] Refactor or cleanup
- [x] Docs or build

## Checklist

- [x] The tracked patch files were updated and the resulting tree passed the build and smoke gates.
- [x] `make build` passed in both Release configurations with 0 errors.
- [x] Every vanilla edit in this follow-up has a `// Stratum` marker.
- [x] No vanilla source is committed.
- [x] `make smoke` reached `WorldReady` and verified 21 console commands.
- [x] `make scenarios` passed 19 of 19 Atlas scenarios.
- [x] The follow-up commit is signed.

## Manual validation

On 2026-09-11, I ran the PR server on a temporary local world with one account connected through RiftLauncher and a second account connected through the standard client. Both accounts authenticated and completed startup.

The moderator account enabled `/vanish`, equipped and mounted a tamed elk, and moved near the observer account. The observer could not see the vanished player in the world or on the map, and `/near` returned no players within 128 blocks. The moderator account could still see the observer, and its `/near` command listed the observer at 4 m. The server audit recorded the elk equipment, mount, and both `/near` commands.

The Atlas reconnect scenario now checks the observer's packet queue for the vanished player's entity ID and also verifies that a plain player's join message remains visible.

## Risk

The change centralizes filtering in existing server-side send paths. Shared mounts with an ordinary passenger remain visible to bystanders and can disclose seat data. Vanilla spectator tracking remains out of scope. Sounds, particles, and block-update side effects remain outside this PR and need separate client-observable validation.


Zaldaryon:
Rebased against `indev` and resolved the conflict in `ServerMain.cs.patch`. #306 and this PR touch different areas of `ServerMain.cs` (inventory packet filtering versus hidden-from-client checks), so I merged the decompiled source directly with `git merge-file` instead of hand-editing the diff text. It merged with zero conflicts, and regenerating every patch from the merged tree produced 190 patches byte-identical to what `indev` already has, plus the combined `ServerMain.cs.patch`.

Build: 0 errors. Smoke: server reached `RunGame`, 21 console commands verified. Atlas: 17/17 scenarios passed in 2m9s.

Head is now `a2b225d`, mergeable against `indev`.

Pixnop (changes requested):
Thanks for doing the merge through `git merge-file` on the decompiled tree. I checked it: applied to vanilla, the head's `ServerMain.cs` differs from indev by exactly the lines e258b61 changed against its old base, and both #306 hunks (`GetPublicStack` in `BroadcastHotbarSlot`, `GetBlockEntityAttributes` in `SendBlockEntity`) are still there.

A good part of #312 is really closed. The thread 0 bucket is filtered where it is built and again in `SendAttributesViaTCP`. Re-evaluating the predicate on every state tick takes care of the hysteresis without a dedicated hook, so losing the vanish privilege or flipping `VanishHideOtherVanishedDefault` converges on the next tick. The filter in `BroadcastEntityPacket` catches every teleport caller at once (/tp, /back, /home, /spawn, jail, respawn, mounted teleport). And the mount branch walks seats the same way vanilla does in `HandleMountPosition`.

Five things need fixing before this goes in.

**1. Arrows still carry the shooter.** The new guard in `SendPrioritySpawn` (and in `PrepareEntitySpawns`) passes the spawned entity to `ShouldHideEntityFromClient`. What goes through `SpawnPriorityEntity` is arrows, thrown spears and fishing bobbers, and the predicate returns false for anything that is neither a vanished player nor a mount carrying one. So a vanished moderator who fires a bow still sends every client in range a spawn at eye height with `firedBy` set to their entity id (`EntityProjectileBase.Initialize` writes it server side before the packet is built), then the arrow's positions. The body lists priority spawns as filtered. Handling `entity is IProjectile { FiredBy: EntityPlayer shooter }` with a vanished shooter the way the mount branch handles a passenger would hide the projectile for its whole life, not only at spawn.

**2. Reconnect restores vanish too late.** `ClearSessionState` still drops the UID from `VanishedPlayerUids` on disconnect, and `RestoreVanishedState` only runs from `OnPlayerJoin`, which fires in `HandleRequestJoin`. By then `SpawnPlayerHere` has already called `SpawnEntity` and sent ServerReady at identification, and the client only sends RequestJoin once it has finished loading. For that whole window, a few seconds and more on a modded server, nearby players get the spawn with exact position and nametag, then position updates, then a despawn. That is #312 item 4, and the body says "restore it before the first join broadcasts". Restoring in the identification path, before `SpawnEntity`, covers reconnects and restarts. Keeping the UID in the set across a disconnect would not cover a restart on its own.

**3. `/vanish off` reveals a player the observers know nothing about.** This one is a regression. Withholding packet 41 (`ToPacketForOtherPlayers`) from hidden viewers in `BroadcastPlayerData` and `SendInitialPlayerDataForOthers` is right. But `RevealPlayerToOthers` and the spawn half of `RefreshVanishedVisibilityForViewer` only send the entity spawn and add the id to `TrackedEntities`, so `UpdateTrackedEntityLists` sees it as already tracked and never takes the path that sends the player data. The vanilla client only fills `PlayersByUid` from that packet. If a moderator was vanished when you joined, or reconnected vanished, and turns it off next to you, you get an `EntityPlayer` with no `ClientPlayer`: nothing in hand, no armor, missing from the player list until they leave range and come back. Sending the player data before the spawn in both paths fixes it.

**4. A shared mount disappears for its own rider.** In the mount branch the first vanished passenger decides, and nothing checks whether the viewer sits on that mount. A regular player steering a two-seat boat or elk with a vanished moderator in the other seat gets their own mount despawned. `UpdateTrackedEntityLists` sends the despawn and the UDP relay stops sending them its position, while the server still has them mounted. Never hiding a mount from a client whose entity is one of its passengers is enough.

**5. The hot path pays for this with nobody vanished.** The predicate used to return right after `entity is not EntityPlayer`. Now every non-player entity goes through `GetInterface<IMountable>()`, which walks the behavior list, for each client in `UpdateTrackedEntityLists`, `SendAttributesViaTCP` and the state tick loop, plus a `LoadedEntities.TryGetValue` in the per-packet position and animation loops. On a server where nobody is vanished, which is nearly always, all of it is wasted. `if (VanishedPlayerUids.Count == 0) return false;` at the top of `ShouldHideEntityFromClient` and `IsVanishVisibilitySubject` makes the common case free, and it matters more once the projectile branch from point 1 is in. CONTRIBUTING asks for before and after numbers on `PhysicsManager`; a rough tick time with 20 players and no vanish would do.

Smaller things, not blocking on their own:

- Join and leave messages still go out for a vanished player: `HandleClientLoaded` only checks `ShowJoins`, `DisconnectPlayer` only `ShowLeaves` and `ShowDisconnects`. Once point 2 restores the state at identification, an `IsVanished` check there closes the rest of #312 item 4.
- The new hunks in both patches carry no `// Stratum` marker (the outer diff adds zero comment lines), and a few rewritten `foreach` headers in `SendAttributesViaTCP` and `SendPositionsAndAnimations` sit one tab deeper than their body.
- The comment above `RefreshVanishedVisibilityForViewer` still points at `PhysicsManager.cs.patch:466` and says it iterates vanished subjects, but it now walks all of `LoadedEntities` and the hysteresis despawns hidden entities by itself. The note on `HideOtherVanishedByUid` sharing a lifetime with `VanishedPlayerUids` is stale too, now that vanish persists.
- `BroadcastHotbarSlot` still goes to everyone. The vanilla client ignores a client id it has no player for, so nothing breaks, but a logging client learns that the vanished player is online and what they hold. The skip list from `BroadcastPlayerData` would cover it.
- The manual pass didn't cover teleport or reconnect, and nothing in `tests/StratumScenarios` touches vanish. Two Atlas players are enough for `/near` and for a reconnect that checks join messages. That last one would fail today, which is a good reason to write it.

Leaving blocks, particles and sounds out of scope is fine by me. #312 asked for that limit to be written down where admins will see it, though, and right now it only lives in the PR body.

Zaldaryon:
Fixed all five blocking points and everything in the smaller list. Head is now `c492b02`.

**1. Arrows.** `ShouldHideEntityFromClient` and `IsVanishVisibilitySubject` now match `entity is IProjectile projectile && projectile.FiredBy is EntityPlayer shooter && IsVanished(shooter.PlayerUID)`, the same way the mount branch handles a passenger. Since the projectile check lives in the shared predicate, it hides the arrow for its whole flight through `SendPrioritySpawn`, `PrepareEntitySpawns`, and the ongoing position and animation paths, not just the initial spawn.

**2. Reconnect timing.** `RestoreVanishedState` now runs in `FinalizePlayerIdentification`, right after `PlayersByUid[playerUID] = client.Player;` and before any of the three `SpawnEntity` call sites that follow (the immediate path, the delayed-chunk `OnLoaded` delegate, and `SpawnPlayerHere`). It passes `persistRevocation: false` there, because the role isn't final yet at that point (`HandleRequestJoin` still upgrades a single-player client to max privilege afterward), and I didn't want to erase a legitimate stored vanish flag against a role that hasn't settled. The existing call from `OnPlayerJoin` keeps `persistRevocation: true`.

**3. Reveal ordering.** `RevealPlayerToOthers` now sends `ToPacketForOtherPlayers` before the entity-spawn packet, and calls `BroadcastHotbarSlot` afterward so the revealed player shows the right held item instead of an empty hand. `RefreshVanishedVisibilityForViewer`'s spawn half does the same through a new `SendPlayerDataForSpawn` helper, since it batches multiple subjects at once, some of which aren't players (mounts, projectiles) and correctly get no player data.

**4. Shared mount.** The mount branch now checks every seat before deciding: if the calling client's own entity is one of the passengers, it returns `false` immediately, regardless of whether another seat holds vanished staff.

**5. Hot path.** Both predicates now start with `if (VanishedPlayerUids.Count == 0) return false;`. I benchmarked with a throwaway Atlas scenario (20 joined players, clustered so everyone tracks everyone, `/stratum timings start/report` over 600 ticks, nobody vanished), run once with the early-out commented out and once with it in place:

| bucket | before | after |
|---|---|---|
| `physics.manager` | avg 0.56ms, max 6.64ms | avg 0.525ms, max 3.70ms |
| `physics.nonTickables` | avg 0.486ms, max 6.61ms | avg 0.491ms, max 3.67ms |

The average delta is small on its own, but the max drops by almost half: the removed `GetInterface<IMountable>()` walk was mostly showing up as tail latency, not steady-state cost, and that only gets worse once the projectile branch runs on every non-vanished tick too. I deleted the benchmark scenario rather than keep a timing-based test in the suite.

**Smaller items**, all done: no join or leave chat message for a vanished player (the leave-message fix reads vanish state before the disconnect system pass runs, since `ClearSessionState` would otherwise have already cleared it); a skip list on `BroadcastHotbarSlot` matching `BroadcastPlayerData`; `// Stratum` markers on every new hunk plus the four one-tab-deep blocks in `SendAttributesViaTCP` and `SendPositionsAndAnimations`; the stale comments above `HideOtherVanishedByUid` and `RefreshVanishedVisibilityForViewer` rewritten to match current behavior; and [`docs/commands/vanish.md`](docs/commands/vanish.md) spelling out that blocks, particles, and sounds stay out of scope.

For the manual-test gap: added `tests/StratumScenarios/VanishPrivacyScenarios.cs` with two scenarios. `Near_Should_HideVanishedStaff_When_CalledByPlainPlayer` runs `/near` through `ChatCommands.ExecuteUnparsed` with an explicit `Caller` (`IWorldSession.ExecuteCommand` runs as console, which would see everything). `VanishedReconnect_Should_NotAnnounceJoin_When_StillVanished` vanishes a player, disconnects and rejoins them under the same name, and checks a second player's actual outbound packet queue for a join-type chat line, since Atlas has no chat sink; it reads the dummy socket's receive buffer directly through reflection and deserializes with `Packet_ServerSerializer`. That scenario fails against the pre-fix code, which is why I wrote it that way rather than asserting on state alone. Both pass now, and the rest of the suite (17 existing scenarios) is unaffected.

Zaldaryon:
## Verification update

I rechecked the current head `c492b022` after the requested privacy fixes.

- `make refresh`: 190 patches applied, 0 failures.
- `make build`: 0 errors in the normal and `EmbedPatchedFiles=true` Release builds.
- `make smoke`: the server reached `WorldReady` and verified 21 console commands without fatal errors.
- `make scenarios`: 19 passed, 0 failed, 0 skipped, in 4m53s.

The requested fixes are already present on this head, so no new commit was needed. Pixnop was requested again through the GitHub reviewer action.

Pixnop (changes requested):
The five blocking points are fixed. I checked each one in the applied code rather than in the patch text, and the choice to put the projectile rule in the shared predicate is the right one: it covers the priority spawn, the spawn batch, entry into range, attributes, positions, animations and tags, not just the initial packet. The restore now runs before all three `SpawnEntity` sites on the identification path, and the player data is already loaded there, so the privilege check reads a real role. The mount branch inspects every seat. The early return makes the common case free.

Five things still need to move before this merges, and none of them is large.

**1. The reconnect scenario does not prove what it says it proves.** The join message is sent from `HandleClientLoaded`, which is packet 26. The event Stratum hangs the restore on fires in `HandleRequestJoin`, which is packet 11, and Atlas sends 11 before 26, exactly like a real client. So under the old placement the vanish flag was already set when the message line was evaluated, and the message was already suppressed. Remove the new call in `FinalizePlayerIdentification`, keep the old one, and this scenario stays green while the leak it claims to cover is fully back. The docstring says the opposite. The scenario is still worth keeping, because it does cover the new message guard, but the claim has to go. The discriminating assertion is cheap now that the outbound queue is already being read: check that no entity spawn carrying the vanished player's id reaches the observer between identification and Playing. That one fails against the old placement.

**2. Two claims in `vanish.md` are not true.** The page says packet 41 hides "the entry in the client's player list". On `/vanish on` no packet 41 is sent at all, and the vanilla client only drops an entry when it receives one with a client id of -99 or below. So an observer who was already connected keeps the vanished player in their online list, while someone who connects afterwards never gets the entry, because the initial send is filtered. The two disagree about who is online, which is itself a vanish signal. Either send the removal packet the disconnect path already builds, or fix the sentence. The closing paragraph is the second one: it tells staff that moving, looking and running commands is safe, and point 4 below is a direct counterexample.

**3. A shared mount vanishes for bystanders, not just for its rider.** The new seat loop returns false only when the viewer's own entity is a passenger. For a third player watching from the shore, the mount is hidden while the ordinary passenger riding it stays visible, so they slide along with nothing under them. That is more conspicuous than the staff member being hidden, and it advertises that someone invisible is aboard. Never hiding a vehicle that carries at least one non vanished player is enough, since hiding the staff entity already does the work.

**4. Death is the third member of the family you just closed.** Join and leave messages are now suppressed for a vanished player, but a death still broadcasts packet 45 to everyone with no skip list, and still posts the death message to general chat through the same `joinLeaveDeathMessage` you guarded twice. A fall, drowning or a wolf is enough to print the name of a vanished moderator, and packet 45 hands a logging client their client id as a join key for everything else. Both fixes have the same shape as the two already written.

**5. The PR body and the suite README are stale.** The body still reports 14 console commands and 17 of 17 scenarios, while the verification comment on this same head says 21 and 19. It says "Addresses #312" where the repo convention is "Fixes #123", so the issue will stay open after the merge. The template sections are gone, which CONTRIBUTING asks not to strip. The README still opens on "Seventeen end-to-end scenarios" and "the eight server boots included", and the measured runtime is now 4m53s.

Outside the diff, the audit turned up four channels that still locate or expose a vanished player. None is a regression from this PR, and I would not hold the PR for them, but they belong in a follow-up issue on #312 rather than being lost:

- `/tpa` is a presence oracle usable by any ordinary player with no modified client. The default survival role carries `stratum.tpa`, the target is resolved by name with no vanish filter, and "No such online player" versus "Teleport request sent" cleanly separates offline from invisible. The vanished staff member also gets a chat notification telling them they were just probed.
- Creatures still target a vanished player in survival. The vanilla rule excludes creative and spectator, and vanish changes neither, so mobs converge on an empty spot and swing at it. That gives continuous position to within about a block while the staff member does nothing at all, and the damage eventually kills them, which lands back on point 4.
- Entity repulsion pushes creatures away from a vanished body. Animals slide around an invisible cylinder, which traces the hitbox.
- The ping packet goes to everyone every five seconds with the full list of client ids and no skip list. A logging client counts the ids it has no player data for. That is the same argument you wrote to justify the hotbar skip list, and it is four lines of the same shape.

One process note: the branch carries a merge of `indev` rather than a rebase, and the comment describes it as a rebase. CONTRIBUTING asks for a rebase. It matters little if this lands as a squash, but the sentence in the thread is inaccurate.

Pixnop (changes requested):
Good pass. I checked the regenerated patch first, because that was the risky part: regenerating `ServerSystemEntitySimulation.cs.patch` moved hunks around the entity throttling blocks. It is clean. I applied both heads onto pristine vanilla and diffed the results: 21 lines differ, all of them the death fix, and both blobs match the post-image hashes in the patch headers. The throttling blocks are byte identical.

Death is properly closed, packet 45 and the message both. The body and the suite README are fixed. The reconnect assertion is real: the spawn packet does reach the observer under the old placement, and the reflected members exist.

One thing needs a decision rather than a patch, and it is my fault for how I phrased the last round.

**The shared mount fix moved the leak instead of closing it.** Making the mount visible when an ordinary player is aboard fixes the gameplay artifact, and that part is right. But a visible mount carries `seatdata` in its watched attributes, and `seatsToAttr` writes `passenger` as each seat occupant's entity id, including the vanished one. A bystander now receives the mount, sits it at the exact position of the vanished staff member, and reads their entity id straight out of the attribute tree. The previous behaviour was not better: hiding the mount left the ordinary passenger floating at that same exact position. Both shapes leak the position, and the current one adds the identity.

So this case cannot be fixed by choosing which entity to hide. Either it gets handled at the root, for instance by refusing to seat a vanished player on a mount that already carries someone else and refusing `/vanish on` while sharing a mount, or it goes into the known limits with the honest sentence: sharing a mount reveals a vanished player. I lean toward the root fix, since a staff member sharing a boat while invisible is already an odd situation, but it is your call and either answer is defensible. What is not tenable is the current `vanish.md` line, which still says a mount is hidden unless the observer rides it. That is no longer what the code does.

Four short corrections:

- **The player list clause is still wrong.** It now reads that an already connected observer may keep the name "until the player leaves range". Leaving range sends an entity despawn, and the vanilla despawn handler clears `worlddata.EntityPlayer` without ever touching `PlayersByUid`. The only thing that removes an entry is a player data packet with a client id of -99 or below, sent on real disconnect. So the entry stays for the whole session. Say that, or send the removal packet on `/vanish on`.
- **No `// Stratum` marker on the new vanilla lines.** The death block adds five lines to a decompiled file and none is marked, while the leave message guard in the same PR is marked correctly. Same for the `deathPacket` and `hiddenPlayers` locals, which CONTRIBUTING asks to prefix.
- **Dead condition in the seat loop.** After the new early return, `vanishedPassenger == null && IsVanished(passenger.PlayerUID)` can only be reached when the passenger is vanished, so the second test never changes the outcome. The comment above the loop also still describes the old rule, which is now the opposite of what the code does.
- **The death message is suppressed globally while the packet is filtered per client.** Packet 45 correctly skips only the clients that must not see the player, but the chat line is dropped for everyone, including staff who can see vanished players. Either is defensible, but the two halves should agree, and neither is mentioned in the documentation, which still lists only join and leave under what vanish hides.

Minor, while you are in there: the docstring now reads "a entity spawn", the body claims a fresh bootstrap applied 190 patches where the repo holds 191, and the template sections are still missing from the body.

One observation on the new scenario rather than a request. It does not pin the two players with `TeleportTo` the way the `/near` one does, so it relies on Atlas dropping both near spawn and on the 128 block tracking range. That holds today. Pinning them would make the guarantee explicit and cost two lines.

Fix the four short items and settle the mount question and I will approve.

Zaldaryon:
## Follow-up on the latest review

Updated the current head `f72f40b6` after the requested changes:

- documented the shared-mount protocol limitation;
- corrected the player-list behavior described in `docs/commands/vanish.md`;
- removed the dead passenger condition from the shared-mount check;
- added the requested `// Stratum` markers around the death and teleport filtering blocks.

## Verification

- `make build`: both Release passes completed with 0 errors.
- `make smoke`: the server reached `WorldReady` and verified 21 console commands.
- `make scenarios`: 19 passed, 0 failed, and 0 skipped in 4m56s.

Pixnop was requested again through GitHub reviewer action. The PR remains blocked while the new head waits for the requested re-review.

Pixnop (changes requested):
The documentation and the predicate are right this time, but this head does not apply. Both modified patches are corrupt, and `git apply --check` rejects them:

```
error: corrupt patch at line 1691   (ServerSystemEntitySimulation.cs.patch)
error: corrupt patch at line 1606   (ServerMain.cs.patch)
```

The same check passes on the previous head, so this commit is what broke them. The cause is that the three comment lines were inserted into the `.patch` files by hand without updating the hunk headers, and a hunk header has to count the lines that follow it:

- `ServerSystemEntitySimulation.cs.patch`, hunk `@@ -384,37 +1753,64 @@` declares 64 post-image lines and now holds 66, because the fix added both the comment and the `// Stratum end` line.
- `ServerMain.cs.patch`, hunk `@@ -3979,15 +4662,30 @@` declares 30 and now holds 31.

Either correct those two numbers, or better, edit the decompiled tree and regenerate with `scripts/extract-patches.sh`, which is what the numbers are for.

Worth knowing why the reported validation did not catch this. Build, smoke and scenarios all run against the working tree, which was already patched before this commit. Editing a `.patch` file does not touch that tree, so everything stayed green while the committed patch stopped applying. Anyone bootstrapping from scratch, including CI, gets the failure instead. A `make refresh` on a clean checkout is the check that would have caught it, and it is worth running before pushing a patch-only change.

One of the three comments is also redundant. The teleport block in `ServerMain.cs.patch` already carried two `// Stratum #312:` lines explaining exactly that, so the added third line repeats them, and it is the one that broke that file. The death block genuinely needed its marker and now has it.

Everything else from the last round checks out:

- The player list sentence is now accurate. It says the entry can survive for the rest of the session because the vanilla client only drops it on a real disconnect, which is what the code does.
- The shared mount limitation is documented honestly, naming both the position and the seat data that discloses the passenger, and the line under what vanish hides no longer contradicts the code. Documenting rather than fixing at the root is a reasonable call, and the Risk section says the same thing.
- The dead condition is gone and the simplification is sound: at that point in the loop the passenger is necessarily vanished, so the behaviour is unchanged for every seat arrangement. The comment above the loop now describes the real rule.
- The body carries the template sections again, with the right counts.

Two leftovers, neither worth another round on its own. The guard around `joinLeaveDeathMessage` still has no marker, which sits awkwardly with the checklist line claiming every vanilla edit in this follow-up has one. And the scenario docstring still reads "a entity spawn".

Fix the hunk counts, confirm `git apply` accepts both patches from a clean tree, and I will approve.

Pixnop (approved):
Approving. The hunk counts are right and I verified the result from a clean tree rather than taking the fix on trust:

- `git apply --check` accepts all three patches against pristine vanilla.
- Every hunk header in the three files now matches the lines that follow it, not just the two that were broken.
- Applying the patches and diffing against the head I reviewed two rounds ago gives exactly three added comment lines and nothing else, so no behaviour moved while the patch text was being repaired.

Across the four rounds this closed the entity, mount, projectile, teleport, player data, hotbar, reconnect, join, leave and death channels, with the vanish state persisted and restored before the first spawn. That is the substance of #312.

Three things to carry forward, none of them blocking this merge.

The `index` lines in both patches are now stale: they still name the post-image from before the comments were added. Nothing reads them today, `git apply` included, so this is harmless, but it is the same signature as the bug we just fixed, namely that the files are being edited by hand. Regenerating through `scripts/extract-patches.sh` updates those lines for free and makes the next contributor's diff clean.

The shared mount sentence in the documentation credits the seat data with too much. The mount stays visible and broadcasts its own position like any other entity, so an ordinary client already sees where the vanished passenger is by watching the boat move. The seat data adds the exact entity id on top of that. The distinction matters because one is visible to anyone on the server and the other needs someone logging packets. While you are there, the page describes the limitation but never tells staff what to do about it, and a line saying not to share a mount while vanished would make it usable as guidance.

The death asymmetry is still open and I am fine with that if it is deliberate. Packet 45 skips only the clients that cannot see the player, while the chat line is dropped for everyone including staff who can see them, which is the opposite of what `/near` and the map pins do. If that is the intended behaviour, say so on the vanish page, which currently lists death messages as hidden without mentioning that staff lose them too.

Left over from earlier rounds, for whenever you next touch these files: the guard around `joinLeaveDeathMessage` still has no marker, and the scenario docstring reads "a entity spawn".

The channels outside this PR that still locate a vanished player, `/tpa` presence probing, mobs targeting, entity repulsion and the ping packet, belong in a follow-up on #312 rather than here.

PR #337, opened 2026-09-19
More precise selection of AVX instructions based on palette size

tehtelev:
## Summary

`ChunkDataLayer.DecodePlanesAvx2` / `DecodePlanesAvx512` (used by `CopyBlocksToUnsafe` and `CopyLightToUnsafe`) previously resolved every palette index through `VPGATHERDD`, regardless of how small the palette actually was. Gather is a comparatively expensive, data-dependent memory operation. Since `palette.Length` is always exactly `2^bitsize` by construction, small palettes fit entirely in one or two vector registers, so the lookup can be done with in-register permute instructions instead of touching memory at all.

This PR adds tiered, capacity-based dispatch to both decode paths: small palettes use permute-based lookup (no gather), larger palettes fall back to the existing gather implementation unchanged. It also adds a dedicated `num == 1` specialization on the AVX-512 side and an ISA-coverage-aware equivalence test.

### AVX-512 path (`DecodePlanesAvx512`)

**Special case selected before the tiers:**

- `num == 1` (`bitsize == 1`) — new `DecodeWordAvx512_Bit1`. Bypasses `BuildIndices512`’s bit-transpose network entirely (designed for up to 8 planes; wasted work when there is exactly one bit of index to extract) and does the whole lookup with `AND` + `CompareNotEqual` + `BlendVariable` on the raw plane word — no permute crossbar at all. This is selected first, and independently of the `palLen`-keyed tiers; it is **not** itself a tier. Correctness relies on the same invariant below (`1 << bitsize <= palette.Length`), which guarantees `palLen >= 2`, so indices 0/1 are always in-bounds and no masking is needed — unlike the general tiers.

**Four `palLen`-keyed tiers, selected once per call:**

- `palLen ≤ 16` — single `VPERMD`-equivalent table lookup (pre-existing `Perm` tier).
- `16 < palLen ≤ 32` — two-table lookup via `VPERMI2D`/`VPERMT2D` (`Perm2`).
- `32 < palLen ≤ 64` — **new**: two `VPERMI2D`/`VPERMT2D` lookups (covering entries 0–31 and 32–63) combined with a `BlendVariable` on index bit 5 (`Perm2x2`).
- `palLen > 64` — unchanged gather fallback.

In all tiers, entries at index ≥ `palLen` are zero in their table slot (tables are zero-initialized and filled only up to `palLen`), so the `idx < palLen ? palette[idx] : 0` contract from the original gather implementation is preserved exactly. The uniform-word fast path (`mixed == 0`) and the tail-word handling (< 1024 words) are untouched.

The dispatch documents the invariant it relies on: every writer keeps `1 << bitsize <= palette.Length` (`MakeSpaceInPalette` doubles `palette` and increments `bitsize` together; `CleanUpPalette` allocates exactly `1 << CalcBitsize(paletteCount+1)`; both `Decompress` paths derive `bitsize` as `floor(log2(palette.Length))`). `BuildIndices512` emits indices bounded by `num`, not `palLen` — the invariant is what makes keying off `palLen` equivalent to keying off `num` for the 5-bit (`Perm2`) and 6-bit (`Perm2x2`) index masks.

### AVX2 path (`DecodePlanesAvx2` / `DecodeWordAvx2_LE8`)

AVX2 has no two-input permute instruction, so each additional tier costs a full extra table plus a deeper blend tree — the tradeoff against gather is worse than on AVX-512. Two tiers were added where the tradeoff is still favorable:

- `num ≤ 3` (`palLen ≤ 8`) — single `VPERMD` table (`tier A`).
- `num == 4` (`palLen == 16`) — two-table `VPERMD` + blend on index bit 3 (`tier B`).
- `num` 5–8 and `num > 8` — unchanged gather fallback.

The palette table(s) are built **once per `DecodePlanesAvx2` call**, not per word, to avoid rebuilding them 1024 times.

### Hardware/runtime notes

- All new branches check `Avx512F.IsSupported` / `Avx2.IsSupported` at runtime, consistent with the rest of the file; no new ISA extensions (GFNI, IFMA, VNNI, etc.) were introduced — those were evaluated and rejected as not applicable to this specific lookup/transpose problem.
- No changes to `palette`/`dataBits` shape, compression format, or `stratumWriteLock`/`readWriteLock` synchronization.

### Testing

Added `ChunkDataLayerDecodeEquivalenceScenarios`: an Atlas scenario that, for bitsize 1–15, generates random / all-zero / all-ones / per-word-uniform plane data, decodes it with a scalar reference implementation, and asserts bit-for-bit agreement against `DecodePlanesAvx2` and `DecodePlanesAvx512` (invoked via reflection). Comparisons for an ISA absent on the running machine are skipped rather than failed, so the test stays meaningful across different CI/dev hardware.

Three additions on top of the original version of this test:

- **Non-power-of-two palette lengths.** Three extra cases per seed (`palLen = 6, bitsize = 2`; `48, 5`; `20, 4`) exercise the `palLen`-keyed tiers with `palette.Length` **not** equal to `1 << bitsize` — i.e. the shapes `Decompress` actually produces, and the only shapes that can falsify (rather than assume) the `1 << bitsize <= palette.Length` invariant the dispatch relies on.
- **ISA coverage reporting.** The run prints which of AVX2 / AVX-512 it actually compared against, and the scenario tracks `avx2Checked` / `avx512Checked` accordingly, so a “quiet green” from a CPU that skipped one of the branches is visible in the log rather than silent.
- **Opt-in strict ISA requirements.** With `STRATUM_REQUIRE_AVX2=1` / `STRATUM_REQUIRE_AVX512=1` set, the test fails loudly if the named ISA was never exercised. This closes the previous hole where a CI leg meant to cover AVX-512 could pass while AVX2 also happened to be available and short-circuited the run. Both vars are opt-in, so a plain dev machine without either ISA still passes with a log line.

The `build` and `smoke-test` commands, as well as all test `scenarios` (21, including the new one), completed successfully.

## Type

- [ ] Bug fix
- [x] Performance
- [ ] New feature
- [ ] Refactor or cleanup
- [ ] Docs or build

## Checklist

## Performance numbers

Seed - seed 1027995113.
World setting - default.
AVX instructions: enabled

**Before**
- `ChunkDataLayer.DecodePlanesAvx512` method execution time: 10 749, 11 486, 11 023 ms.
- **Average:** 11 086,0 ms
- **Standard deviation:** 372,52 ms

**After**
- `ChunkDataLayer.DecodePlanesAvx512` method execution time: 8162, 8696, 8421, 8867, 8717 ms.
- **Average:** 8 572,6 ms
- **Standard deviation:** 280,32 ms

The average execution time of the method decreased by **22,67%** (from **11 086,0 ms** to **8 572,6 ms**).

**Before**
- `ChunkDataLayer.DecodePlanesAvx2` method execution time: 12 604, 12 185, 13 460 ms.
- **Average:** 12 749,67 ms
- **Standard deviation:** 649,86 ms

**After**
- `ChunkDataLayer.DecodePlanesAvx2` method execution time: 11 470, 11 270, 11 651 ms.
- **Average:** 11 463,67 ms
- **Standard deviation:** 190,58 ms

The average execution time of the method decreased by **10,09%** (from **12 749,67 ms** to **11 463,67 ms**).

## Bonus 
Bitsize histogram
image
This distribution reveals further potential options for improving the algorithm.
I think this can be continued in the next PR if a noticeable gain is demonstrated.


Pixnop (dismissed):
Approved. I went straight at the one thing here that could have been serious, and it holds.

The new tiers choose `Perm2` and `Perm2x2` from `palLen`, while the indices they resolve are bounded by `num`. A palette shorter than `1 << num` would therefore wrap indices into occupied slots instead of yielding 0, and that is world data on its way to clients. It cannot happen: every path that sets `palette` and `bitsize` keeps `1 << bitsize <= palette.Length`. `MakeSpaceInPalette` doubles the array and increments `bitsize`, `CleanUpPalette` allocates exactly `1 << num8`, and both decompression paths derive `bitsize` as `floor(log2(palette.Length))`. So `perm2` implies `num <= 5` and `perm2x2` implies `num <= 6`, and the 5-bit and 6-bit masks are exact. Worth one line of comment next to the tier flags recording that the tiers rest on that invariant, because the code reads as if it were keying on the wrong quantity.

Checks on my side: hunk counters all match, the patch applies to a pristine baseline, `extract-patches` round trips it byte for byte, the build is green, and `make scenarios` gives 21/21 on a CPU carrying both AVX2 and AVX-512, so neither `IsSupported` guard skipped.

Comments, none of them blocking:

- The new scenario builds every case with `paletteLength = 1 << bitsize`, which is the one shape where the tiers cannot be wrong. The interesting shapes come out of `Decompress`, where `palette.Length` is whatever the saved data carried: length 48 with `bitsize` 5 lands in `Perm2x2` with indices that only reach 31, length 6 with `bitsize` 2 lands in `Perm`. A few non-power-of-two lengths would test the invariant above instead of assuming it.
- On a machine without AVX-512 the whole AVX-512 half is skipped and the run still reports green, because `anyIsaChecked` is already true from the AVX2 branch. Nothing runs the suite automatically, so that is a silent hole sitting exactly where this PR spends its risk. Either report which ISAs actually ran in the test output, or fail when the AVX-512 comparison was skipped.
- `ScalarDecode` describes itself as an exact copy of the scalar fallback, but the shipped one reads `plane != null ? plane[word] : 0` and the oracle reads `planes[k][word]` unguarded. Null planes are the one input where the two could genuinely disagree, and the oracle cannot express it.
- `Assert.True(false, message)` at line 93 is the only new warning in the build (xUnit2020). `Assert.Fail(message)` is the drop-in.
- Both decoders end on `if (word < 1024)` after a loop stepping by two from zero, so those tails never execute. The AVX-512 one is the only route from that path into `DecodeWordAvx2`, which leaves dead code sitting on the ISA boundary this PR rearranges, and you already touch the AVX2 one.
- The before and after numbers are not interleaved and use different sample counts. The AVX-512 gain is comfortably outside the spread; the AVX2 one at about 10 percent is roughly twice the quoted deviation, so it reads as suggestive rather than settled. Interleaving a handful of runs would close that.

Thanks for the scenario. A decode equivalence test against a scalar reference is what this area was missing.

tehtelev:
I implemented items 1, 2, 3, and 5 (invariant comments, dead tail assertions, `Assert.Fail`, ISA coverage reporting, and non-power-of-two `palLen` cases). I intentionally skipped the null-plane case.

I verified the invariant before adding it: a null plane within the `0..num-1` range (i.e., `dataBits[k]` where `k < bitsize`) represents an unreachable state rather than an edge case requiring test coverage. `LoadWordsAndClassify`/`LoadWordsAndClassify2` explicitly document this as a precondition ("Guaranteed that plane != null for k < num"), and the code paths that could potentially violate this do not do so: `CleanUpPalette` sets `dataBits[i] = null` only for `i >= num8` (i.e., `i >= bitsize` after compression); `SetBlocksBulkUnsafe`/`SetLightBulkUnsafe` clear the entire state before rebuilding; and palette expansion (`MakeSpaceInPalette`, `PaletteIndexForBulk`) always allocates a new plane immediately before increasing `bitsize`, ensuring it never remains null within the range the decoder is actively reading.

Adding such a `DecodeCase` would test for a violated invariant that shouldn't exist—rather than protection against a real-world scenario—and would very likely simply crash with an NRE in the SIMD paths (which lack the guard present in the scalar fallback), resulting in a false signal rather than a genuine issue discovery.

The `build` and `smoke-test` commands, as well as all test `scenarios` (21, including the new one), completed successfully.

tehtelev:
A `bitsize=1` branch has been added to the AVX-512 path. A similar trick did not yield a noticeable improvement in the AVX2 path.
Test measurements have been updated.

Pixnop (approved):
Approved again, on 82b7b51.

Five of the six comments are handled, and handled properly. The non-power-of-two cases are in with the exact shapes that matter (6 with 2 bits, 20 with 4, 48 with 5, one per tier). The run now says which instruction sets it actually compared, and on this machine it prints `AVX2=True, AVX-512=True`. The oracle reads a null plane as 0 like the shipped fallback, `Assert.Fail` replaced the constant assert, and the three unreachable tails are gone. The invariant comment at the dispatch is what I hoped for.

The new one-bit tier is correct. Broadcasting each word, masking per lane against the two bit-position vectors, comparing to zero and blending `palette[0]` with `palette[1]` is the scalar definition for `num == 1`, and the invariant guarantees the palette has at least two entries, so the read of `palette[1]` is always valid.

Checked here: 26 hunks out of 26 consistent, the patch applies to a pristine baseline, the build is green with no new warning on the file, and `make scenarios` is 21/21.

Three notes, none of them blocking:

- `Debug.Assert(word == 1024, ...)` compiles to nothing in the Release configuration, which is what ships and what `make build` produces. It still documents the intent well, just do not count on it firing.
- `STRATUM_REQUIRE_AVX512` and `STRATUM_REQUIRE_AVX2` are not set anywhere in the Makefile or the scripts, and the coverage line only shows at detailed verbosity. A sentence in the suite README telling people how to turn a skipped instruction set into a failure would make that work pay off.
- `LoadWordsBit1` reads `planes[0]` unguarded. That matches every other SIMD loader, which all rest on the "plane != null for k < num" precondition while only the scalar fallback checks. It is not something this PR introduced. If the null case ever needs covering, the place is one check at the dispatch that falls back to scalar, as a separate change.

The AVX2 numbers are as before, within about twice the spread, and that is fine: the AVX-512 side carries the case on its own.

